PR #3, opened 2026-06-10
gamepad

IoIxD:
## Summary

Adds basic gamepad support, optimized for an Xbox 360+ Controller. DPad is used for navigating, Start is mapped to what Return does, Back is mapped to Escape. Perhaps this could work with an actual remote? I dunno.

There's no support for rebinding because I'm not good with QML and am not at liberty to decide how that'd be put in a settings menu. Also you have to configure mpv seperately for gamepad support.

The way it's done is admittedly jank, I looked into how you're supposed to do it and I'm not sure what QtQuick wants you to do to be able to listen on gamepad events? So we initialize SDL and then synthesizes key events.

## AI involvement

If I had the money for Claude Code I'd maybe consider using it? But alas, I coded this myself.

## Testing

- Platform(s) tested: x86_64 Linux
- Display / output tested: Main monitor

## Checklist

- [?] Sizing and Positioning with `root.sh` / `root.sw` (no hardcoded pixels) and mindful of CRT overscan
- [x] No tracking/analytics; only writes to the local data directory

(there's no root.sh file? and no there's no tracking or analytics)


anthonycaccese:
thank you @IoIxD! I'll look to review and test your changes in the next few days.  

For gamepad rebinding, I started thinking about handling that through a config file in the configuration directory.  Thinking something like...

```
// input.cfg
SDL_CONTROLLER_BUTTON_A [ACTION]
SDL_CONTROLLER_BUTTON_B [ACTION]
SDL_CONTROLLER_BUTTON_DPAD_LEFT [ACTION]
SDL_CONTROLLER_BUTTON_DPAD_RIGHT [ACTION]
SDL_CONTROLLER_BUTTON_DPAD_UP [ACTION]
SDL_CONTROLLER_BUTTON_DPAD_DOWN [ACTION]
...
```

If that file exists then use those mappings and if not fall back to defaults (and I agree with you that xbox 360 is a good baseline for the defaults).  

And for [ACTION] defining a set of core navigation and mpv actions that those could map too.  I would likely look to refactor how i am handling keyboard and usb remote mapping now too to make like a central "input" manager vs how I am doing it in the QML right now if I go that route.  

But wanted to share my initial thinking to see what you thought, thank you for any feedback.

And regarding: `there's no root.sh file? and no there's no tracking or analytics`
I'm sorry about not making those clearer in the docs and PR template:
- `root.sh` in this context refers to the QML side of things where root refers to a root sizing value and .sh refers to "screen height".  its used to calculate vertical sizing and spacing through percentages so that 240-MP scales to both 4:3 and 16:9 (or other aspect ratios).  So yes for your changes for gamepad you were spot on, you didn't need to use that all =)
- and for tracking and analytics, I'll make it clearer that it means... please don't add any tracking or analytics =).  Basically I want to make sure that any additions to 240-MP remain privacy focused (no tracking of how users are using it, no signals reported out, etc...).  My thinking is it should just act like a VCR you would have bought in the 80s/90s.  You plugged it in, it played things and never kept or reported any data on that action.   So you're all good. 

I'll make those clearer in the PR template instructions, thank you for raising them.

anthonycaccese:
Hi @IoIxD, thank you again for taking a pass at this, it helped me sit and think about the approach I wanted to take more and made it a real idea instead of just a task on my list to think about =)

With that, I took the path of building a centralized input manager to sit in the app core leveraging SDL.  I'll push up my branch with that approach soon and add it here for you to check out. 

It supports a few things I was wanting to include with gamepad support: hot swapping, help system awareness (changing the labels based on the active input), and input/label overrides through an input.cfg to account for the "is A or B the right button for select" question =)

anthonycaccese:
Ok opened a draft PR here: https://github.com/anthonycaccese/240-MP/pull/15

anthonycaccese:
Thank you again for proposing this, with the changes added in https://github.com/anthonycaccese/240-MP/pull/15 I'll close this in place of those updates.

PR #19, opened 2026-06-13
Allow right shift key to go back (like ESC) for one handed keyboard use

tmcleroy:
## Summary

It allows the right shift key to be used to navigate back (like ESC) for one handed (right handed) keyboard use.

## AI involvement

All code and comments were AI generated by Claude Fable 5. A human reviewed the code and manually tested.

As called out in the comments, right shift during video playback does not work to go back in the Mac build due to the MPV issue but does work in the Pi build.

## Testing

Manual human testing was done on both the Mac and pi (I used a pi 3b+) builds. The mac build was tested on an M4 macbook air display and the pi build was tested on a CRT although there is no visual change introduced by this PR.

## Checklist

### DIFF
--- a/src/input/InputManager.cpp
+++ b/src/input/InputManager.cpp
@@ -488,6 +488,11 @@ bool InputManager::eventFilter(QObject *obj, QEvent *event) {
     // gamepad Back path, which posts Escape into QML — or sends ESC to mpv
     // over IPC when fullscreen mpv holds OS focus and the window can't take
     // key events. The bare Shift event is consumed; no view binds Key_Shift.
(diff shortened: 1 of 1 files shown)


anthonycaccese:
thank you @tmcleroy ❤️, this is a nice quality of life change and I agree the small gap on MPV for MacOS is an acceptable.  I'll add a note to my list to explore that later as part of future input work.  I confirmed on MacOS and RPi as well and its a nice addition.  Will merge this now and will make time to review you're other PRs after breakfast.  Cheers again.

PR #20, opened 2026-06-13
fix(plex): persist audio/subtitle selection and use a fresh session per play

tmcleroy:
## Problem
On the Plex detail view's **PLAYBACK SETTINGS**, the chosen **AUDIO**/**SUBTITLES** were not reliably applied to playback, and the selection reset to OFF when returning to the screen.

Root cause: the picks were never sent to Plex. `set_audio_stream`/`set_subtitle_stream` existed but were never called, so:
- returning to the detail screen reloaded Plex's stored default (looked like a reset to OFF), and
- transcodes could burn Plex's default streams instead of the chosen ones (the server selects from its stored default).

## Fix
On play (`Item.qml` `Keys.onReturnPressed`):
- Persist the selected audio/subtitle to the part via `set_audio_stream`/`set_subtitle_stream` (`subtitleStreamID=0` = off).
- Regenerate the playback `sessionId` per play so Plex builds a fresh transcode for the current selection instead of reusing a stale one.

QML-only change.

## Testing (human)
Tested on **macOS** and **Raspberry Pi 3B+**, in both **direct play** and **transcode** modes. Selection now persists across returning to the detail screen; direct play applies the correct audio/subtitle tracks. Transcode is substantially improved (some Plex server-side edge cases may remain — tracked separately).

## Notes
- Persisting selection sets the part's default **server-side**, so the choice also sticks in other Plex clients (standard Plex behavior).
- Not addressed here: direct-play selection of downloadable **text/SRT** subtitles (separate mpv `--sid` mapping issue, rare), and residual **transcode** edge cases (Plex-side caching/burn timing) — follow-up investigation noted.

## AI disclosure
Implemented with AI assistance (Claude Code); diagnosis and code reviewed and hardware-tested by a human.

### DIFF
--- a/modules/plex/views/Item.qml
+++ b/modules/plex/views/Item.qml
@@ -22,8 +22,12 @@ FocusScope {
     property int audioIdx: 0
     property int subtitleIdx: 0
 
-    // Unique session ID for this playback instance
-    readonly property string sessionId: {
+    // Session ID for the current playback instance. Regenerated on every play
+    // (see Keys.onReturnPressed): reusing one lets Plex hand back a stale
+    // transcode session built with the previously selected audio/subtitle.
+    property string sessionId: newSessionId()
+
+    function newSessionId() {
         var chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
         var id = ""
         for (var i = 0; i < 12; i++) id += chars[Math.floor(Math.random() * chars.length)]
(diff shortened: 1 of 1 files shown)


anthonycaccese:
thank you @tmcleroy, great find on the transcode + subtitle use case and great fix for the subtitle persist.  looks good and testing well, adds nice QOL with the persist option for sure.

Side note: if you have any interest... these were some of the other plex things I was thinking to explore:
- Explore optimized version playback 
    - if a given item has optimized versions available then offer them as options for playback
    - could be useful for RPi where there may be 480 optimized versions already encoded to avoid the need for transcoding
- Explore Up Next support for TV Episodes
    - I wasn't sure about adding this because i like that the UI creates focused action that asks you to select the next thing to play but figured it still might be worthwhile to explore
- Music Library support - started a discussion where I am thinking through ideas here: https://github.com/anthonycaccese/240-MP/discussions/26 

If any of those sound like things you are interested in i'd be glad to collaborate and no worries if not

Just hope you are still having fun and that 240-MP is working well for your setup - thank you for your contributions

PR #21, opened 2026-06-13
Switch Plex servers from the main menu

tmcleroy:
## What

Adds a way to quickly switch the active Plex server from the module's main menu, instead of having to go back to Settings. The Plex module connects to the default/configured server on entry; this lets you jump to any of your other configured servers in a couple of keypresses.

## How

- **`Libraries.qml` (main menu):** press **◄/►** to open the server picker. A footer hint (`[◄►]:SERVER`) appears only when more than one server is available. Left/Right were unused on this vertical list and are delivered identically on gamepad (d-pad → arrow keys), so this works on keyboard and controller with no `InputManager` changes.
- **`PlexBackend::get_switchable_servers()`:** new synchronous `Q_INVOKABLE` returning the active user's servers as `{name, machineId}`, filtered the same way as `getServers()` (managed users only see their own servers). Reads from cached auth, so no network round-trip.
- **`ServerSelect.qml`:** reuses the existing picker. A new `switching` nav param makes it return to the refreshed main menu on success (the initial auth flow is unchanged) and pre-highlights the currently active server.

After selecting a server, `select_server` persists it and the menu reloads with that server's libraries. Cancelling returns to the menu, and the nav stack stays clean (Back from the menu still exits the module).

## Testing

Human-tested: verified Plex server switching works end-to-end on both macOS and Raspberry Pi 3B+.

### DIFF
--- a/modules/plex/views/Libraries.qml
+++ b/modules/plex/views/Libraries.qml
@@ -14,6 +14,10 @@ FocusScope {
     property var libraries: []
     property string serverName: ""
     property string userName: ""
+    // Servers the active user can switch to from the main menu. More than one
+    // means the quick-switch action (◄/►) is offered.
+    property var switchableServers: []
+    property bool canSwitchServer: switchableServers.length > 1
 
     Connections {
         target: plexBackend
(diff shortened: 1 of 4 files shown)


anthonycaccese:
This is a very nice addition plus I like the UI pattern you implemented of "<>" being a way to get to the server selection list, very nice.  Tested auto sign in on/off and with a managed user use case to make sure it only displays servers shared with them vs all the servers shared with an admin and it.  All worked perfectly.  Great work and I very much appreciate the focused scope of work and clean implementation.  Cheers @tmcleroy

PR #36, opened 2026-06-15
feat: plex module - add loading states when launching playback

tmcleroy:
## Summary

Playing a movie could appear frozen while Plex prepares the stream or while mpv starts up — after pressing PLAY nothing changed on screen for a while. This adds loading feedback covering both gaps, reusing the Plex module's existing `LOADING...` pattern (no new component).

## Changes

- **`Item.qml`** — show a full-screen `LOADING...` overlay (with a `BACK:CANCEL` hint) the moment PLAY is pressed, until we navigate to the player or hit a backend error. Nav keys are ignored mid-launch to avoid a double-launch / stream-switch race.
- **`Player.qml`** — show `LOADING...` on the black screen while mpv launches and buffers, cleared on the first position update. Also covers the transcode re-request and the direct-play→transcode failure retry. Stays hidden under the resume prompt.
- **`Player.qml` (Pi fix)** — starting mpv runs synchronously and, on the Pi, switches VT (suspending Qt's render thread) before the `LOADING` frame can paint. So only the async "start from beginning" transcode path (which yields to the event loop) showed loading. The synchronous launch paths (RESUME, the `Component.onCompleted` direct-play branch, and `startFromBeginning`'s non-transcode branch) now defer one tick via a short `Timer` so the loading frame renders before the mpv handoff. macOS is unaffected (no VT switch; the defer is imperceptible).

## Testing

Human-tested on both **macOS** and a **Raspberry Pi 3B+**:
- Item screen: `LOADING...` appears immediately on PLAY and `BACK` cancels it.
- Player: loading text shows during mpv startup and disappears once playback begins; not shown under the resume dialog.
- Confirmed RESUME (and direct play) now show the loading screen on the Pi, matching START FROM BEGINNING.

### DIFF
--- a/modules/plex/views/Player.qml
+++ b/modules/plex/views/Player.qml
@@ -26,6 +26,7 @@ FocusScope {
     property string selectedSubtitleId: navParams.selectedSubtitleId || "0"
 
     property bool stoppedReported:    false
+    property bool playbackStarted:    false
     property bool overlayVisible:     false
     property int  choiceIndex:        0
     property string resumeSetting:    "ask"
@@ -203,7 +204,12 @@ FocusScope {
         target: mpvController
 
         function onPositionChanged(ms) {
-            if (ms > 0) playerRoot.lastKnownPositionMs = playerRoot.absPos(ms)
+            if (ms > 0) {
+                playerRoot.lastKnownPositionMs = playerRoot.absPos(ms)
+                // First position update means mpv is up and playing — drop the
+                // loading indicator (mpv's own window now covers the screen).
+                playerRoot.playbackStarted = true
+            }
         }
         function onDurationChanged(ms) {
             if (ms > 0) playerRoot.lastKnownDurationMs = ms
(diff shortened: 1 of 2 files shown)


tmcleroy:
@anthonycaccese I went with a full screen "LOADING..." UI that I've seen elsewhere in this project (see screenshot). Let me know if you want to take a different visual approach, I'm happy to amend my PR.

Screenshot 2026-06-15 at 12 47 29 AM

modules/plex/views/Player.qml
```diff
@@ -255,13 +279,25 @@ FocusScope {
...
     Rectangle {
         anchors.fill: parent
         color: "black"
+
+        // Shown while mpv launches and buffers the stream (before its window
+        // takes over). Hidden once the first position update arrives, or while
+        // the resume prompt is up.
+        Text {
+            text: "LOADING..."
+            color: root.tertiaryColor
```
anthonycaccese: what do you think of setting this to `"white"` instead of `root.tertiaryColor`?

I like both so I am not sure, I mainly started thinking white because it matches with the MPV overlay when that is displayed so it might help tie the loading indicator visually into the playback experience.

anthonycaccese:
@tmcleroy great addition!  I got so used to looking at the blank screen during testing while my plex server spun up from idle that I totally missed this as an idea 😄 great work.

Visual approach looks good, nice reuse of the loading pattern from other views for sure.  Just added a review comment on color to get your feedback on.

Otherwise I'll do a review and test tonight after work.

anthonycaccese:
This is a great QOL addition @tmcleroy, I tested more tonight and I like this addition quite a lot.  Great work.  Only open question I think I have is on the color for the loading indicator.  Using it more this evening I am thinking "white" to match the MPV overlay colors.  I am glad to make that change after merge but just wanted to see your take on it before I do.  I appreciate you for working on this and your other PRs

tmcleroy:
@anthonycaccese good suggestion, implemented in my latest commit. I agree white fits more with the MPV overlay and is more high contrast and readable on CRT especially.

PR #40, opened 2026-06-16
Fix: allow rewinding past the resume point on transcoded playback

tmcleroy:
## Bug

When you resume a video partway through (RESUME from a saved `viewOffset`), you couldn't rewind or seek to any point *before* the resume position. The video's timeline effectively began where you resumed — scrubbing back past that point did nothing. This only affected **transcoded** playback; direct play was fine.

## Cause

For a resume, the transcode session was requested with the resume offset baked in (`request_transcode(..., viewOffset)`). Plex's HLS stream then starts at that offset, so mpv only ever receives the portion of the video from the resume point onward — there is literally nothing earlier in the stream to seek back to. (Direct play was unaffected because mpv loads the whole file and just uses `--start`.)

## Fix

Always request the transcode at offset `0` (full timeline) and have the Player seek mpv to the resume position via `--start` instead. The entire video before the resume point stays in the stream, so rewinding works. `fastSeek=1` is already set on the transcode request, so Plex seeks to the resume position rather than transcoding linearly up to it.

- `Item.qml`: request the transcode at offset 0 instead of `viewOffset`.
- `Player.qml`: `doStartPlayback` seeks mpv to the offset for the transcode path (previously it started the offset-baked stream at 0). The direct-play→transcode fallback gets the same treatment, and `startFromBeginning` simplifies to a plain start-at-0 (no re-request needed).
- Because the transcoded stream's timeline now matches the video's, the position-reporting adjustment (`transcodeStartOffset` / `absPos`) and the `pendingZeroStart` re-request path are no longer needed and were removed. Rebased on top of the autoplay-next-episode change; that feature's paths are preserved.

## Testing

Human-tested on a Raspberry Pi 3B+ driving a CRT, and on the macOS build: resuming a transcoded video and rewinding to before the resume point now works, and start-from-beginning / direct play / the direct-play→transcode fallback all behave correctly.

### DIFF
--- a/modules/plex/views/Item.qml
+++ b/modules/plex/views/Item.qml
@@ -171,7 +171,10 @@ FocusScope {
             sessionId = newSessionId()
 
             if (detail.forceTranscode) {
-                plexBackend.request_transcode(detail.ratingKey, detail.partKey, sessionId, audioId, subId, detail.viewOffset || 0)
+                // Always transcode from the start so the full timeline is seekable.
+                // The Player resumes by seeking mpv to viewOffset (see doStartPlayback),
+                // which lets the user rewind past the resume point.
+                plexBackend.request_transcode(detail.ratingKey, detail.partKey, sessionId, audioId, subId, 0)
             } else {
                 plexBackend.build_stream_url(detail.ratingKey, detail.partKey, sessionId)
             }
--- a/modules/plex/views/Player.qml
+++ b/modules/plex/views/Player.qml
@@ -33,7 +33,6 @@ FocusScope {
     property bool overlayVisible:     false
     property int  choiceIndex:        0
     property string resumeSetting:    "ask"
-    property bool pendingZeroStart:   false
     property bool pendingRetryTranscode: false
 
     // Autoplay-next-episode. When enabled, a natural end-of-file advances to the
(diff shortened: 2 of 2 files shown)


tmcleroy:
@anthonycaccese let me know if you see any issues with this approach, you know the codebase better than I do of course.

You can verify that the bug existed before this change by resuming any video and being unable to rewind past the start point (when transcoding is enabled)

anthonycaccese:
thank you, i'll review and test after work tonight

anthonycaccese:
Sorry for my delay @tmcleroy, I had time to review and test tonight and it works great.  Requesting the transcode at offset 0 and seeking mpv via `--start` is a nice approach to handling this i think.  I had a ton of trouble working through transcoding in my first pass and I really like how your approach has fewer moving parts now, its much cleaner so thank you.

The main thing I'd ask for your help on is there is a merge conflict on player.qml and i wanted to see if you can help to address that conflict on your PR branch before merge.  Its two small things: startFromBeginning() and onPositionChanged (items that were updated from your work on the loading indicator)

The main thing to watch for looks to be in startFromBeginning().  For that I think you'd want to keep the collapsed single-path version but call `beginPlayback(0)` instead of `doStartPlayback(0)` (that should keep the loading indicator in place for "Start from beginning"). 

And for onPositionChanged, keep = ms (drop absPos) but keep your playbackStarted = true.  Let me know if that makes sense and thank you again for your work on this.

tmcleroy:
@anthonycaccese I rebased on latest main and fixed the conflicts. Both as you suggested: startFromBeginning() is now the single collapsed path but calls beginPlayback(0) so the loading indicator still shows, and onPositionChanged keeps = ms (dropped absPos) while keeping playbackStarted = true. Heads up it was a force-push. Builds clean, should be conflict-free now. Tested manually again, works fine

anthonycaccese:
thank you @tmcleroy! with this change in, i'll prep a release for tomorrow.

PR #51, opened 2026-06-18
feat(plex): auto sign in when only one user is registered

tmcleroy:
## What

When entering the Plex module with exactly **one** registered user, skip the user-selection screen and sign that user in automatically — showing a brief "Signing In" state instead of presenting a list with a single choice.

Previously the module always showed a full "Select User" screen even when there was only one account to choose.

## How

In `modules/plex/views/UserSelect.qml`:
- Extracted the selection logic into a `selectUser(user)` helper (handles both the reauth and first-login paths).
- In `Component.onCompleted`, after loading users from cache, if `users.length === 1` it sets `autoSelecting = true` and signs that user in directly.
- The `autoSelecting` state swaps the UI: AppBar subtitle reads "Signing In", a centered "SIGNING IN..." indicator shows, and the user list + footer hints are hidden (matching the existing `LOADING...` pattern in `Libraries.qml`).

### Back-stack fix
When auto-selecting, the view advances with `replaceWith` rather than `navigateTo`, keeping it **out of the back stack**. Otherwise pressing back from the main screen would land on UserSelect, which would immediately re-trigger sign-in and bounce the user forward again.

## Coverage

Both entry paths are handled:
- **Reauth** (auto-sign-in off) → auto-selects → `Libraries`
- **First login** (`needs_user`) → auto-selects → `ServerSelect`
- **Multi-user** → unchanged, manual selection as before

## Testing

Manually verified on macOS, and on a Raspberry Pi 3B+ on a CRT: a single-user account enters Plex straight into the library, and pressing back from the main screen exits the module cleanly instead of looping through sign-in.

### DIFF
--- a/modules/plex/views/UserSelect.qml
+++ b/modules/plex/views/UserSelect.qml
@@ -7,9 +7,34 @@ FocusScope {
     property var navParams: ({})
 
     signal navigateTo(string path, var params)
+    signal replaceWith(string path, var params)
     signal goBack()
 
     property var users: navParams.users || []
+
+    // True when there is exactly one user and we are signing in automatically,
+    // skipping the manual selection step entirely.
+    property bool autoSelecting: false
+
+    function selectUser(user) {
+        if (!user) return
+        if (navParams.reauth) {
+            plexBackend.reauth_select_user(user.id)
+        } else {
+            plexBackend.select_user(user.id)
+        }
+    }
+
+    // Advance to the next view. When auto-signing in the only user, replace
+    // rather than push so this screen stays out of the back stack — otherwise
+    // pressing back from the next screen would land here and re-trigger sign-in.
+    function goForward(path, params) {
+        if (autoSelecting) {
+            userSelectRoot.replaceWith(path, params)
+        } else {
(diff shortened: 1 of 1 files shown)


tmcleroy:
@anthonycaccese FYI I did not test this with an instance that has multiple plex accounts as I only have one, it works fine for the single plex user case though. Do you have multiple plex accounts? If not I can create another for testing purposes tomorrow.

anthonycaccese:
thank you @tmcleroy, the module has a setting for auto sign in that I think should handle the use case directly.  

Check out the `Component.onCompleted` check on line 89 in `modules/plex/views/Root.qml`

Can you try setting Auto Sign On to `On` in settings > plex and see if that works for you? (I have multiple users, have that set to on and when opening the module Root handles the routing for me to Libraries.qml currently)

Wondering if maybe I missed something when an account only has one user but hoping the auto sign on setting already handles this.

tmcleroy:
@anthonycaccese sorry about that I didn't notice that setting. I tried it out and it does handle the use case. I'll close the PR

anthonycaccese:
all good and no worries!  thank you for raising it.

PR #55, opened 2026-06-19
fix(local): Add auto subtitles setting

danhunsaker:
## Summary

This adds a setting to automatically show subtitles (when enabled) for the local player.

It also includes a small change to the subtitle logic in `MpvController.cpp`. It now assumes/asserts that subtitles should always be disabled when `subTrack` is negative, but an empty `subFiles` list doesn't mean there are no embedded subtitles available/desired, nor that MPV shouldn't look for local subtitles on its own.

Together, these changes allow improved QOL for those with hearing disorders (deaf, hard of hearing, auditory processing disorder, etc) or in settings where sound is off, but dialogue is still desired.

## AI involvement

No AI used.

## Testing

- Platform(s) tested: Pi 5 (no Mac, sorry, and my other Pi boards were thefted a few years back :( )

Steps to test:
- Started a video known to contain embedded subtitles and confirmed they were not automatically displayed
- Stopped the video, toggled the setting, and resumed playback, confirming the subtitles are now, indeed, shown automatically
- Just for good measure, stopped playback and toggled the setting off again, then resumed playback once more to confirm the setting is still honored when toggled off manually. Just in case.

All tests worked as expected!

## Checklist

### DIFF
--- a/modules/local_files/manifest.json
+++ b/modules/local_files/manifest.json
@@ -38,7 +38,7 @@
     },
     {
       "key": "auto_subtitles",
-      "label": "Automatically Show Subtitles",
+      "label": "Auto Show Subtitles",
       "type": "toggle",
       "default": "OFF"
     }
--- a/modules/local_files/manifest.json
+++ b/modules/local_files/manifest.json
@@ -35,6 +35,12 @@
       "options_source": "dynamic",
       "options_slot": "get_resume_playback_options",
       "default": "ask"
+    },
+    {
+      "key": "auto_subtitles",
+      "label": "Automatically Show Subtitles",
+      "type": "toggle",
+      "default": "OFF"
     }
   ]
 }
--- a/modules/local_files/views/Player.qml
+++ b/modules/local_files/views/Player.qml
@@ -18,6 +18,8 @@ FocusScope {
     property bool   shuffleOn:           false
     property string resumeSetting:       "ask"
 
+    property bool   subtitlesOn:       false
+
     // Track last non-null values during playback for robust save on exit
     property int    lastKnownPositionMs:  0
     property int    lastKnownDurationMs:  0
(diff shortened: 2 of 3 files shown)


anthonycaccese:
thank you @danhunsaker, i'll aim to do a review and test through this weekend

anthonycaccese:
@danhunsaker thank you for this PR, def agree its a great QOL update.

I made 2 commits for your review 2 were minor style edits (alignment of a property and renmaing of the label to "Auto" to align the the language used for other auto settings)

And suggested removing the call to subtitlesOn within Keys.onPressed to match how loopOn / shuffleOn settings are handled, they are read once at playback start and not re-read per key.

Let me know your feedback on those updates, glad to talk them through.

Outside of that, I tested with a few use cases and things look good:
- External srt file 
- Embeded sub track
- Forced srt + non forced srt

Appreciate you suggesting and doing the work to add this.  Cheers.

danhunsaker:
I suppose I should have looked a bit closer at when `Component.onCompleted` gets called. 🤦  I'd interpreted it as only firing when a video stops playing, for some reason.

Everything looks great to me!

Biduleman:
It would be nice to have a way to set the language we want for the default sub in a "best effort" way. So if you add the "eng" language then if the "eng" track exists you get that one, otherwise you get the first one.

Also, it would be nice to have a bit more logic when disabling subtitles. [forced] subtitles are usually wanted even when subtitles are disabled as they're there for a specific reason, so I feel they should stay enabled, or at least have an option to keep them enabled even if the subtitles are disabled.

danhunsaker:
I was planning to tackle those bits in a separate PR. Without keyboard support, it gets complicated supporting all the myriad languages out there in a configuration menu, so I wanted to split that out to keep this one simple enough to merge quickly/cleanly.

For the meantime, `~/.config/mpv/mpv.conf` supports these settings in a set-and-forget capacity.

anthonycaccese:
I agree with both of you.  this is a good focused PR for a starting point and maybe one (or 2) in the future to consider forced on by default or preferred subtitle language order could be interesting.  mpv.conf is a perfect place for those now.

PR #57, opened 2026-06-19
feat: smooth-playback toggle to restore crop on Pi 3

tmcleroy:
## Problem

On the Pi 3, the OSC **CROP** button blanks the screen instead of cropping (reproduces on real hardware/CRT, not on desktop).

Root cause: `e6ae377` ("mpv playback improvements") switched the Pi 3 to a zero-copy DRM-**overlay** decode path (`--vo=gpu --gpu-context=drm --hwdec=v4l2m2m`) for smoother playback. That plane can't zoom/scale, so mpv's `--panscan` (what CROP triggers) produces a black frame. Before that commit the Pi 3 used `--vo=drm --hwdec=auto-safe`, which goes through the normal scaler — so crop worked. Crop capability was the (documented) trade-off of the new path.

## Change

Adds a user-facing **Smooth Playback** setting (default **On**) so the user can choose:

- **On** (default) — keeps the smooth zero-copy overlay path; CROP can't zoom.
- **Off** — falls back to the crop-capable scaler path (`--vo=drm --hwdec=auto-safe`) at higher CPU / slightly less smooth cadence.

Details:
- App-level `smooth_playback` setting; read when mpv launches, so it applies to the next video.
- `MpvController::hasSmoothPlaybackTradeoff()` gates the toggle to the only profile where the conflict exists (Pi 3). On Pi 4 / Pi 5 / desktop crop already works, so the toggle is hidden there.
- Settings shows the toggle with a help line noting it disables cropping. Adds a generic per-row `description` field to the Application section for this.

## Testing

- **Manually tested on a Raspberry Pi 3B+**: with Smooth Playback **Off**, the CROP button zooms correctly instead of blanking; with it **On**, playback uses the smooth overlay path as before.
- **On macOS the Smooth Playback setting does not appear** — this is expected: the toggle is gated to the Pi 3 profile via `hasSmoothPlaybackTradeoff()`, since every other profile can already crop and the setting would be a no-op there.

### DIFF
--- a/src/player/MpvController.cpp
+++ b/src/player/MpvController.cpp
@@ -528,7 +528,7 @@ void MpvController::appendVideoArgs(QStringList &args) const {
             if (smoothPlaybackEnabled())
                 args << "--vo=gpu" << "--gpu-context=drm" << "--hwdec=v4l2m2m";
             else
-                args << "--vo=drm" << "--hwdec=auto-safe";
+                args << "--vo=drm" << "--hwdec=v4l2m2m-copy";
         } else {
             // Pi 5 (Full KMS) and the safe fallback for unknown headless Linux.
             args << "--vo=drm" << "--hwdec=auto-safe";
--- a/src/player/MpvController.cpp
+++ b/src/player/MpvController.cpp
@@ -522,7 +522,13 @@ void MpvController::appendVideoArgs(QStringList &args) const {
             // straight to a DRM overlay plane for the lowest possible CPU (~15%) with
             // smooth playback. The one trade-off: the overlay plane can't zoom/crop,
             // so mpv's --panscan (the OSC crop button) blanks the video on this path.
-            args << "--vo=gpu" << "--gpu-context=drm" << "--hwdec=v4l2m2m";
+            // The "smooth_playback" setting (default ON) lets the user opt out: when
+            // OFF we fall back to the crop-capable scaler path (--vo=drm) at the cost
+            // of higher CPU and less smooth cadence.
+            if (smoothPlaybackEnabled())
+                args << "--vo=gpu" << "--gpu-context=drm" << "--hwdec=v4l2m2m";
+            else
+                args << "--vo=drm" << "--hwdec=auto-safe";
         } else {
             // Pi 5 (Full KMS) and the safe fallback for unknown headless Linux.
             args << "--vo=drm" << "--hwdec=auto-safe";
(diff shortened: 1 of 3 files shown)


tmcleroy:
@anthonycaccese after submitting this PR I just noticed in the [BUILDING.md](https://github.com/anthonycaccese/240-MP/pull/49/changes#diff-f1355fd54329147c830843ab133fed7716180c3242031487b861a2a3235d0078) of the PR that introduced smooth playback that it is possible to override it with a config file change.

It's unfortunate that smooth playback breaks the crop feature on the pi3. I think we would both agree that users need an option to enable crop at the expense of smooth playback, but I'm not sure what the best approach is (config file edit vs settings page toggle). I can see pros and cons to both approaches, neither is ideal.

Editing a config file to restore crop functionality is significant friction to less technical users and users who aren't aware that it's necessary may wonder why the crop button blacks out the screen on their 3b+ (this is what I was wondering when I discovered it this morning, lol)

On the other hand I get why we wouldn't want to clog the settings page with another setting to restore crop at the expense of smooth playback (although it is only visible to pi3 users)

For my personal use case since I'm outputting composite video to a CRT I transcode everything to 480p anyway and don't worry about smooth playback as even the pi3 can handle 480p just fine. And crop functionality is essential because even 4:3 content still often shows black bars on my CRT.

What are your thoughts on config file vs settings toggle?

anthonycaccese:
I had a big internal debate about this when making the change so I def understand your POV @tmcleroy 😀

My initial rationale for not adding a setting toggle for it was because it was a Pi3 only concern and I wasn't sure of a clean way to implement a device specific way to approach that toggle.   So I opted for the config route.

So I thought about the approach you've taken in this PR a lot today though and I think it fits well.  Primarily that it enables an on device way for Pi3 users to handle their preference.

I tested it on Pi4 and Pi3 to make sure the new setting still allows `mpv_video_args` to be set in the config and that if the "smooth_playback" setting exists in a pi4 configuration it doesn't impact Pi 4 at all.  Your guard for that  by only including it under Pi3 video playback on line 519 in src/player/MpvController.cpp does the trick nicely I think.

I initially liked the idea that you tried for adding a help note pattern in settings but I think long term as we add additional modules the layout may start to feel a bit overwhelming, so I pushed a change back to settings.qml that removes it and does some other tweaks.  

The main tweak being changing the label from Smooth Playback to 1080p Playback - thinking with that label change it makes it easier to make the decision to remove the help text as the label indicates exactly what it does. 

Interested in your feedback on that change so please let me know.

I'd def be down to revisit the help text idea in the future, i'd just want to rethink the overall UI for the settings I think.

anthonycaccese:
Sitting with this a bit more, I def see the value of the help description... specifically in this case it does feel like the best to way to express what the toggle does AND its caveat (something the label itself won't convey alone).   So I'll play with the UI a bit more tomorrow to get it back in - will push a change for your to check out soon.

anthonycaccese:
Played with the UI this morning, trying the accent color as a background to try to align it visually with the currently selected row and added help text for color scheme.  Also worked with the layout to fit it between what would be the last row of the list and the footer with some margins to align things.

Please let me know what you think @tmcleroy, if this works well then we can use it as a help text pattern on other views when it makes sense I think.

Photos:

tmcleroy:
@anthonycaccese looks good to me, much more readable 🚀

anthonycaccese:
r2BtghAUTmpP2

Sweet, will merge now - great work on this @tmcleroy

PR #63, opened 2026-06-22
feat: local files module - add subtitle language preference

danhunsaker:
## Summary

Getting this working was, as expected, a bit more involved than the last couple of changes I submitted.

- ~~First, an extension of `MultiSelectSettings` to add support for a slot for default values, along with a switch for whether options should default to selected/on (as previously, and still the default) or unselected/off.~~
- Second, slot logic in `LocalFilesBackend` to look up the languages ~~realistically supported by most systems (based on what Qt itself provides), alongside the list of system locales (also provided by Qt) for use as defaults (while also boosting them to the top of the options list, since they'll be most likely to be looked for)~~ recognized by ISO 639-1 from a JSON file. Configuration in `manifest.json` wires this new logic into the `local_files` UI.
- ~~Third, fallback-font-loading logic in `main` coupled with fonts (under `assets/fonts` as expected) which contain the additional scripts and characters VCR_OSD_MONO leaves out, starting with a monospace version of Unifont (UnifontExMono), then falling back to the non-mono *official* Unifont's two parts (basic plane and upper plane(s)) for glyphs still not found - Unifont is designed to include a glyph for every single printable character in Unicode itself, and is up to date as of Unicode 17 (UnifontExMono is only up to Unicode 16 at the moment, which is why we can fall back from it as well).~~
- Finally, tying it all together and making it actually useful, changes in the `local_files` `Player` and `MpvController` to look up and actually *apply* the setting by passing `--slang` to `mpv`. (Empty `[]` arguments were also added for the other modules' `Player`s to properly support the added argument.)

Possible improvement, but complex enough to be beyond scope for what I'm wanting to focus on at the moment, would be to rank the subtitle languages rather than simply select them, since an order would be helpful in cases where users know multiple languages and have a preference for which order to try them.

## AI involvement

No AI used, except the now-unavoidable AI summaries when doing web searches to look things up.

## Testing

- Platform(s) tested: Pi5
- Video with multiple subtitle tracks in multiple languages, to test switching between them via settings changes

## Checklist


anthonycaccese:
thank you @danhunsaker, I'll dig into this after work tomorrow for feedback.

anthonycaccese:
I am sorry I missed on my original timing for review, I guess the good news though is having time to review and think through this PR alongside your other for https://github.com/anthonycaccese/240-MP/pull/66 got me thinking about maybe a different approach for this one.

Thinking since one of the goals is to enable a user to set their primary preferred language so that when a video loads it preselects a subtitle with that language... that maybe we simplify this to just a single select "preferred subtitle language" instead.  

Using the multiselect to choose multiple languages worked but it felt complicated in practice deciding on the full list of languages I may want to "allow".  Where if I was able to just set a preferred language then that could default easily and still allow me to toggle to other languages available in the stream if i wanted to

Another way I guess I am thinking about this is that we shouldn't try to "limit" the available streams in a file but instead just try to default to a preference from that file (hope that makes sense)

So if this changed to a single select "preferred subtitle language" setting then we could maybe think of the functionality like this:
- If the setting is set to a specific language then when the video loads we try to find a subtitle in that language and default to it only IF...
    - Auto Subtitle is set to "On"
    - OR
    - Auto Subtitle is set to "Forced" (in this case also try to specifically load a subtile in that preferred langauge with the forced flag on it)

Def down to discuss this though, still thinking through it as you can likely tell =)

danhunsaker:
I could see either option.

- Single select is a lower cognitive load, for sure
- Multiselect (for subtitles) is supported by MPV (and similar) mostly so that if the first language isn't available, you at least get *something* you can understand from the very first onscreen text (unless, of course, *none* of your preferences are available)
- My main hesitation with the current multiselect is still that we can't really rank these in a preference order, which would generally be preferred over Qt's automatic alphabetic sort - addressing that would certainly need to be either a different `manifest` `type` (`ranked_submenu`?) which stores an array instead of an object in the config, or an additional option for `multiselect_submenu` to switch implementations (which can either store an array as above, or simply store `int` rather than `bool` and sort things on use).

I'm not sure what kind of UI would work best for ranking things like this, which is why I didn't add it already (though I don't have a good reason for not starting a Discussion about it 😅). I was imagining `On` being replaced by 1-through-number-`!Off`, but I'm sure there are other, and potentially better, approaches.

anthonycaccese:
re: "Single select is a lower cognitive load, for sure" and "My main hesitation with the current multiselect is still that we can't really rank these in a preference order, which would generally be preferred over Qt's automatic alphabetic sort"
- what do you think of trying the [ISO 639 language codes](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) as the root of the list instead of QTs list?  Its structured and is aimed to cover all known natural languages and looking at other services that provide a similar setting (like Plex for example) they base from the ISO codes.  That could reduce the cognitive load I think?
- then try that list as a single select for now (smaller list to switch through) and with the pattern added for list_single to allow for display/value (https://github.com/anthonycaccese/240-MP/pull/72) it could support leveraging the ISO codes for functionality with the display of the setting being the full language name.

I did start thinking of a ranking UI today that could be used for things like ranking the order of modules on the homescreen.  Its kind of similar to the UI in place for library view where you can press "right" to get to a secondary nav item in the list

The rough idea:
| Select the Item to Reorder | Press right to select the move option | press up/down to reorder |
| --- | --- | --- |
|  |  |  | 

That said I am thinking for this case that single select might be way to go, single select your preferred language to set the default essentially.

danhunsaker:
I grabbed Qt's list because I really didn't want to hardcode my own, and pulling in dependencies is something I try to avoid, but I'll make these changes in just a bit.

anthonycaccese:
thank you @danhunsaker

danhunsaker:
🤦🏻 Of course I missed one...

anthonycaccese:
Thank you @danhunsaker and I apologize for the time delay on my end thinking through this one.  I just tested your updates and they are working well from my end.  Ran through a set of test cases (not explicitly set, set to any and set to ES and EN with test srt files to confirm) and all worked really well.

I just made one small edit to subLangString to fall back to "-" for sub_lang to cover both the case for when its unset and when its explicitly set to "any" so it will work the same for both cases (i.e. not set a --slang preference).  Before that when it was unset it was falling back to "any" which ultimately was ok as it would't find a match but figured to get them set up the same would be easier to follow in the long run.

Thank you for your prep for future array handling across the modules too.  Will commit this now and the follow with the forced option PR.

anthonycaccese:
you are all good! just came up in testing and I figured it was just a simple miss (and your original intent) so I added it

PR #64, opened 2026-06-22
feat: add Jellyfin module

### LINKED ISSUE #47: [Idea]:  Jellyfin Module - First Pass
### Area

A new module

Create a module to support a direct integration with Jellyfin

Discussion and scope occurring here: https://github.com/anthonycaccese/240-MP/discussions/4

PR with work done by @RP2 started here: https://github.com/anthonycaccese/240-MP/pull/64

### Why is this useful?

Would provide a way for users to view content from their Jellyfin server through 240-MP.  Provides a simple VCR experience for focused viewing on a CRT.

RP2:
## Summary

Add a Jellyfin module for browsing and playing media from a Jellyfin server, including Quick Connect authentication, HLS transcode playback, library browsing (movies, TV shows, seasons, episodes), continue watching, resume support, and autoplay next episode.

Also includes cross-cutting improvements: API token redaction in mpv logs, `--ytdl=no` for all playback (moved outside Plex-only block), and Release-build debug stripping.

SSL verification is relaxed for the configured Jellyfin server host only (self-signed LAN certs)

## AI involvement

This work was developed using opencode go models (DeepSeek V4 Flash/Pro and Kimi K2.7). I reviewed the generated code to the best of my ability, but my background is web development rather than C++/Qt and QML. I manually tested feature additions to catch issues with UX/UI and video playback.

## Testing

- Platform(s) tested: Arch Linux (primary), Raspberry Pi 3b+ (some), Raspberry Pi 5 (Composite to CRT)
- Display / output tested: Standard HDMI display and CRT TV, both worked

## Checklist

Closes #47


anthonycaccese:
AWESOME @RP2!  I was just making progress on a version of this and then I saw your PR ❤️.  So i'll start reviewing and adding feedback here this week after work each night, thank you so much for building this.

src/player/MpvController.cpp
```diff
@@ -151,9 +152,13 @@ void MpvController::loadAndPlay(const QString &url, float startSeconds,
     {
         QFile lf(m_logFilePath);
         if (lf.open(QFile::Append | QFile::Text)) {
+            QString safeUrl = url;
+            safeUrl.replace(QRegularExpression("[Aa][Pp][Ii][_-]?[Kk][Ee][Yy]=[^&\\s]+"), "ApiKey=REDACTED");
```
danhunsaker: Might consider:

```cpp
safeUrl.replace(QRegularExpression("Api[_-]?Key=[^&\\s]+", QRegularExpression::CaseInsensitiveOption), "ApiKey=REDACTED");
```

src/player/MpvController.cpp
```diff
@@ -201,11 +206,11 @@ void MpvController::loadAndPlay(const QString &url, float startSeconds,
         args << QStringLiteral("--shuffle");
     if (muteAudio)
         args << QStringLiteral("--no-audio");
+    // yt-dlp hook intercepts HTTP media URLs and can break Plex/Jellyfin
+    // playback with spurious 401/400 errors — disable unconditionally.
+    args << QStringLiteral("--ytdl=no");
```
danhunsaker, line 248: This will break some other (future?) use cases that *aren't* Plex or Jellyfin - would be safer to check for tokens for both instead of assuming nothing legitimate will *ever* want `ytdlp` support.

src/player/MpvController.cpp
```diff
@@ -315,7 +320,12 @@ void MpvController::loadAndPlay(const QString &url, float startSeconds,
         // approach doesn't apply here; --osd-fonts-dir is provider-independent.
         args << QString("--osd-fonts-dir=%1").arg(m_appRoot + "/assets/fonts");
 #endif
-        qDebug("[MpvController] desktop launch: mpv %s", qPrintable(args.join(" ")));
+        QString safeCmd = args.join(" ");
+        // Redact all token forms in debug output
+        safeCmd.replace(QRegularExpression("[Aa][Pp][Ii][_-]?[Kk][Ee][Yy]=[^&\\s]+"), "ApiKey=REDACTED");
```
danhunsaker: `QRegularExpression::CaseInsensitiveOption` would help here, too.

src/player/MpvController.cpp
```diff
@@ -222,7 +227,7 @@ void MpvController::loadAndPlay(const QString &url, float startSeconds,
     connect(m_process, &QProcess::readyRead, this, [this]() {
         const QByteArray out = m_process->readAll();
         if (!out.isEmpty())
-            qWarning("[mpv] %s", out.trimmed().constData());
+            qDebug("[mpv] %s", out.trimmed().constData());
```
danhunsaker: Curious why the logging level change, here...

views/ModuleList.qml
```diff
@@ -137,7 +137,7 @@ FocusScope {
         
         Keys.onReturnPressed: {
             var selectedModulePath = menuList.model[menuList.currentIndex].entry_point
-            console.log("Routing to: " + selectedModulePath)
+            // [dev] console.log("Routing to: " + selectedModulePath)
```
danhunsaker: Not sure why this was turned off...

modules/jellyfin/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Launches mpv subprocess. The `video_quality` setting controls the transcode resolution:
+- **Transcode**: `master.m3u8` — HLS stream, server-side decode + re-encode
+
+mpv receives `--ytdl=no` unconditionally to prevent yt-dlp from intercepting media URLs.
+
+## Pi / CRT notes
+
+- Pi 3B+ uses `--hwdec=v4l2m2m` (zero-copy) for smooth playback. This path cannot crop — widescreen content gets black bars on 4:3 CRTs.  
+  To enable cropping: override `mpv_video_args` in `config.json` (e.g. `"--vo=gpu --gpu-context=drm --hwdec=v4l2m2m-copy --panscan=1.0"`). Copy mode uses more CPU but panscan works.
```
danhunsaker: This was just changed in `v2026.06.22` (see #65) to be toggleable from Settings. That said, your suggested args here aren't quite the same as the ones added by #57, so the difference might be of interest to @tmcleroy and/or @anthonycaccese to compare.

src/modules/jellyfin/JellyfinBackend.cpp
```diff
@@ -0,0 +1,1150 @@
...
+#include <QSet>
+#include <QRegularExpression>
+
+static const QString kModuleId = QStringLiteral("com.240mp.jellyfin");
+
+static QString authHeaderValue(const QString &token, const QString &deviceId) {
+    QString auth = QStringLiteral("MediaBrowser Client=\"240-MP\", Device=\"%1\", DeviceId=\"%2\", Version=\"%3\"")
+                       .arg(QSysInfo::machineHostName(), deviceId, QCoreApplication::applicationVersion());
+    if (!token.isEmpty())
+        auth += QStringLiteral(", Token=\"") + token + QStringLiteral("\"");
```
danhunsaker:
```cpp
        auth += QStringLiteral(", Token=\"%1\"").arg(token);
```
for consistency/clarity.

danhunsaker (commented):
These thoughts are mine, so nothing official. Many are questions rather than suggestions. Looks mostly awesome, though!

RP2:
Thanks for the feedback @danhunsaker, I appreciate you taking the time to go through this.
A few thoughts on the outstanding items before I get to the change log:
- --ytdl=no unconditionally – All four modules (Plex, Jellyfin, Local Files, Ambient Mode) pass direct media URLs to mpv. None of them need yt-dlp (no YouTube/streaming-site content), and the Plex module had it scoped to Plex-only since the initial release. Since every current playback path is unaffected, it's simpler to set it once in MpvController rather than duplicating per module. Happy to revisit if a future module actually needs yt-dlp.
- qWarning → qDebug – With QT_NO_DEBUG_OUTPUT in Release builds, mpv stderr noise logged via qDebug is stripped entirely (no disk/CPU waste on Pi). qWarning would still emit, and mpv's stderr is chatty, most of it isn't actionable. This way dev builds see it, Release builds don't.
- Commented-out console.log in ModuleList.qml – That was a leftover dev trace. Commenting it rather than removing keeps the routing line handy for future debugging without cluttering the terminal during normal use. No strong feelings either way though, I can drop it entirely if you prefer.

Latest round of changes:
- Resume positions now save correctly – PlayMethod was set to DirectPlay while the actual stream was always HLS transcode. Jellyfin uses PlayMethod to validate session state, so stop reports (and their PositionTicks) were silently ignored. Changed to Transcode across all three reporting endpoints, added StartPositionTicks for accurate start tracking, and a Failed flag on error exits so corrupt positions don't get saved.
- Video quality actually enforces resolution – The setting was only sending MaxStreamingBitrate, but the server doesn't downscale if the source fits within that bitrate. Added MaxHeight to PlaybackInfo to force a proper transcode to the selected tier.
- --monitorpixelaspect=0.82 – Corrects non-square pixel aspect on Pi 5 composite out (704x432 on a 4:3 CRT). Zero-cost math, makes a big difference on CRTs.
- Playback stop signal robustness – Added failed flag handling in the onPlaybackEnded handler and a Component.onDestruction safety net in Player.qml so Jellyfin always receives the stop signal regardless of how playback ends.
- ApiKey regex cleanup (per your suggestion) – Switched to CaseInsensitiveOption + Api[_-]?Key pattern in both log-file and safeCmd redaction. Also tidied up auth header style with .arg().
- Migrated to unified playbackEnded signal – Upstream refactored three exit signals into one. Jellyfin player now branches on reason to decide autoplay vs exit.

src/player/MpvController.cpp
```diff
@@ -534,7 +544,10 @@ void MpvController::appendVideoArgs(QStringList &args) const {
                 args << "--vo=drm" << "--hwdec=v4l2m2m-copy";
         } else {
             // Pi 5 (Full KMS) and the safe fallback for unknown headless Linux.
-            args << "--vo=drm" << "--hwdec=auto-safe";
+            // --monitorpixelaspect=0.82 corrects non-square pixel aspect on
+            // composite CRTs (704×432 on 4:3). Pure math, zero rendering cost.
+            args << "--vo=drm" << "--hwdec=auto-safe"
+                 << "--monitorpixelaspect=0.82";
```
anthonycaccese, line 585: May I have the background on the change to add this hardcoded pixel value?  Where does the math come from? How was it tested on a CRT to confirm?

views/ModuleList.qml
anthonycaccese: +1 same question for me

src/player/MpvController.cpp
anthonycaccese: +1 same question for me

anthonycaccese:
Just writing to say i'll get the time to focus and run through this in full over the weekend, I thought I would be able to spend more time during the week but since this is the first new module as a contribution i'd like to be able to spend more dedicated time running through it.

src/player/MpvController.cpp
RP2: The Pi's composite output runs at 704×432 pixels. On a 4:3 CRT, mpv assumes square pixels and treats that as a ~1.63:1 display, which is wider than the actual 4:3 screen. The result is 4:3 content gets rendered too narrow, with black bars on the sides.

pixel_aspect = (4/3) / (704/432) = 0.82

This corrects the pixel aspect so the image fills the screen properly. I verified it by playing 4:3 content on a Pi 5 connected to a CRT via composite to make sure the bars went away and geometry looked correct. The bug I brought up in #73 is closely related to this.

One thing to keep in mind: this lives on the Pi 5 DRM path which covers both composite and HDMI. On HDMI, mpv reads physical dimensions from EDID and should ignore the override, but if anyone hits issues they can adjust it through the existing mpv_video_args config setting.

RP2:
> Just writing to say i'll get the time to focus and run through this in full over the weekend, I thought I would be able to spend more time during the week but since this is the first new module as a contribution i'd like to be able to spend more dedicated time running through it.

Hey @anthonycaccese I appreciate you taking the time to give it a proper look! There's no rush at all, take your time. I might push a couple more small fixes in the meantime.

src/player/MpvController.cpp
RP2: Most of what mpv writes to stderr isn't actually warnings but rather stream info, codec init, etc. qDebug keeps them visible during dev and automatically silenced in Release builds. Happy to revert if that's preferred.

views/ModuleList.qml
RP2: Just a debug line I commented out so it didn't log routing info in production, can re-enable if desired.

anthonycaccese:
ok, i've been testing and working on things this afternoon, i'm running through a final set of tests on an update to playback that adds direct play as an optiom, refactors the transcode logic a bit to align it up with some of the learning we had in the plex module and a fix for subtitle display.  

Direct play was the big one I think and works quite well on the Raspberry Pi for h264 (and HEVC on the pi4 and above) so thought it would be good addition plus it aligns with what we have in place for Plex as well.  

Ok will wrap up testing soon and make a commit, following that I am going to look at adding the "browse by letter" functionality we have in plex to the library views on jellyfin - that will help with browse speed in larger libraries.  And then i'll do another scrub of testing through end to end.  

I wrote up some notes on auth that we can revisit in future work too - specifically thinking of a set of UI paths we can add to enable multiple users for a given server as well as a way for user to add additional servers (both of those are nice to haves from what folks have said so far so not critical path for right now I think)

Overall though nice work @RP2 - my hugest thanks for kicking this off.   And if you have any concerns with me making updates please let me know, will be glad to talk it through.

anthonycaccese:
ok @RP2 I made a bunch of good progress today so I am closing things for the night, the module is getting to pretty close parity in terms of experience to the plex module now.  testing is looking good with my test library and on a RPi + CRT.

tomorrow I have a few more things i'd like to look at:
- Auto play episode takes a long time to start the next episode (it does work but just takes time)
- RSHIT was added as an alternative back button for one handed use which impacts using that to type special characters into the server url field.  Will look to address this tomorrow

Longer term i'd like to take a look at auth a bit more things like:
- local server probe as an alternative for having to type out the server url
- adding a way to quick connect for a different user on the same server (to allow for user switching)
- Collections and Genre browsing (need to see what Jellyfin supports)

When you get time to test please let me know any feedback and hope these updates work well on your end.

RP2:
Hey @anthonycaccese I appreciate the work today!

One quick question: the kSupportedCollectionTypes filter in JellyfinBackend.cpp drops boxsets (Collections) from the library list. I'm assuming this is intended since your inline comment mentions mirroring plex, but could boxsets be added back to the supported types? I personally just find it an easy way to quickly jump into a series and it seems like it's also something you might like to see long term.

Besides that, I found one bug I missed earlier regarding resume position. I'll continue to poke around to see if I find anything else.

anthonycaccese:
most def we can add collections (boxsets)!

i just didn't have any set up to test how they work so i opted for including known tested flows first (I am still learning how jellyfin works) - i'll get a few set up tomorrow to play with the user flow as it looks like a collection can include and display movies and shows together.

Also thinking to check out playlists

src/player/MpvController.cpp
anthonycaccese: I had time to spend with this today and its a good call to flag but approach wise we can look to handle this in a module that needs ytdl either directly in its backend or a kind of per-playback opt-in flag as needed (similar to how we are handling other args now).   So I am ok with removing it like this.  I'll start researching a youtube module soon I think and then I can get into real testing of the idea there.  

Thank you for raising this and the review, ok to keep this change

anthonycaccese:
ok @RP2 spent some time building, testing and watching today and made a few additional changes that I think could work well.

- on RPi fixed a control quirk where when pressing back with the OSC open it would stop playback instead of just closing the OSC
- added a global fix to allow RShift to be used in textinput fields, so now when you press RShift to add a : in the server url it won't back you out a screen =) - this change also refactored the UI on both auth and quickconnect a bit to align to the ui patterns in other modules.  but glad to talk through them if you prefer to revert to another style.
- I saw your note about keeping deviceID in the _auth file even after sign out but in practice I didn't find that it made a difference to keep the ID or not so I opted to make it work like Plex where a signout fully removes the auth details.  In practice signing out removes the device from auth on the server end anyway so I wasn't sure keeping a stale ID was needed in that process considering the deviceID is randomly generated.  But glad to talk through that change.
- reverted the logging things that were raised as part of the initial review (I find those details helpful to have even in release builds so opted to keep them in)

Have watched a few movies and episodes while working through things just to get a feel of how its working overall and I think its working pretty well end to end.

Long term I can see the potential ask for multi user and multi server but figured those can come later if they come up as  requests - for now as an initial release the single server/user setup is working nicely I think

I'll take time to dig into the collections use case tonight - I have an idea on how to approach it where it can support the display of movies, shows and collections (I found testing today that collections can contain collections!) so will make a focused commit for that later if I get something working well.

Other than that would love your feedback on how things have been working in your use/testing of everything too.  If you are feeling good I think getting this merged in soon could work well.

RP2:
Thanks @anthonycaccese, this looks good.

- OSC close fix: clean, good fix.
- RShift + auth UI refactor: textInputHasFocus() via Qt::ImEnabled is the right approach. The UI alignment with the rest of the app is nice. I also appreciate the right shift fix as well since it's what I use out of muscle memory.
- Signout clears auth: I agree, stale deviceID serves no purpose after server-side revocation. Generating a fresh UUID is correct.
- Logging: agree, it's still useful during active dev.

I believe this is ready to ship. Collections and multi-user can come in focused follow-ups.

One thing I did catch: modules/jellyfin/views/Item.qml was missing the navListState property declaration that every other view has, which causes a ReferenceError when returning from the player. I've added it.

anthonycaccese:
Have boxsets in a good state I think, it covers that a boxset can contain any number of sub types inside of it (movies, series, episodes and other boxsets) as well as will auto collapse to a single list if it contains only one subtype (like a boxset that only contains movies)

also wires it up so it shows as a library type that can be turned on/off within 240-MP (which is how jellyfin handles it too - as a library type)

here is an example end to end showing how it works in my test set up after running the TMDB boxset plugin

Settings:
Screenshot 2026-06-28 at 8 07 27 PM

Libraries:
Screenshot 2026-06-28 at 8 07 36 PM

Collections:
Screenshot 2026-06-28 at 8 08 14 PM

A Collection with both Movies and Series in it:
Screenshot 2026-06-28 at 8 08 22 PM

Selected Movies:
Screenshot 2026-06-28 at 8 08 31 PM

Selected a Movie:
Screenshot 2026-06-28 at 8 08 38 PM

Collection name persists through the navlist and backstacking works all the way back up,

Logically I think this all works well and should cover the use cases I understand from building a few collections this weekend so fingers crossed once its out there for real world use =)

anthonycaccese:
awesome was a good day and i'm closing out for the night =)

I have things pretty much staged for a merge and will likely do that tomorrow after some additional testing.  Collections was a good win and getting homevideos wired up correctly wraps things nicely too.  

If you see anything odd in testing just shout and add fixes as you see them.  We can either do things before a merge or in a follow up PR before a release (which we can likely do this week as well) - all good either way.

anthonycaccese:
Added 2 things found during testing the last few hours, first was an update to the subtitle logic to respect the subtitle mode preference set on the server (so it more cleanly handles the default subtitle to select on item.qml based on server preferences - gives one source of truth for that now which I think will be better in the long term)

anthonycaccese:
ok merging now and closing this PR with the following details:

feat: add jellyfin module

Only supports single server and single user for this initial pass (supports quickconnect auth for user)

Features for this initial pass:
- Designed for CRT navigation (simple, fast, list browsing)
- Supported library types: `movies, tvshows, homevideos, boxsets`
- "Quick Connect" authentication
- Select specific libraries to display
- Continue Watching, Next Up and Resume Playback support
- Autoplay next episode in a season (optional, off by default)
- Collections support
- Select preferred audio/subtitle track before playback and switch tracks during playback
- Subtitle selection infomred by the server+user's subtitle mode
- Full library browsing by letter
- Show/Season browsing
- Video quality selection: Direct Playback (Default) or Transcode options

thank you again @RP2 for kicking this off and your work on it.

anthonycaccese:
wiki added: https://github.com/anthonycaccese/240-MP/wiki/Module:-Jellyfin

PR #66, opened 2026-06-23
feat: local files module - upgrade auto subs to include forced-only option

danhunsaker:
## Summary

A few changes to make this work:

- Update `manifest.json` to use a `list_single` instead of a simple `toggle`, and `forced` instead of `off` as the default
- Update `LocalFilesBackend` to provide the updated options list ("Forced Only", in addition to the previous "On" and "Off"), complete with info on how to handle the old values in existing configs (though "Off" in the old config is mapped to "Forced Only" in the new, since this is what most folks actually want in the first place)
- Update `Player` to support both the old `boolean` and the new `string` values
- Update `ModuleSettings` to also support `old` settings alongside current ones, for any/all `list_single` settings that define them (currently just `auto_subtitles`, but flexible for any future settings that may also need to become `list_single` from something else)
- Update `MpvController` to treat `subTrack == -1` as forced-only, and `subTrack < -1` as off - all other values retain their previous meanings, though I noticed subtitles wouldn't load without a language selected in the `mpv.conf`, so I did update the `subTrack == 0` logic to pass two additional flags to correct that

Much simpler changes than the ones to add language selection support, but I'll definitely need to do some merge conflict resolution on whichever of the two gets merged second. :sweat-smile:

## AI involvement

No AI used

## Testing

- Platform(s) tested: Pi 5
- Display / output tested: HDMI (and RustDesk remote access, but that's functionally the same thing)

## Checklist

### DIFF
--- a/modules/local_files/manifest.json
+++ b/modules/local_files/manifest.json
@@ -39,8 +39,11 @@
     {
       "key": "auto_subtitles",
       "label": "Auto Show Subtitles",
-      "type": "toggle",
-      "default": "OFF"
+      "type": "list_single",
+      "options_source": "dynamic",
+      "options_slot": "get_auto_subtitles_options",
+      "default": "forced",
+      "description": "Whether to automatically show subtitles when a video starts.\n[FORCED ONLY] displays for foreign dialogue only, when available."
     },
     {
       "key": "sub_lang",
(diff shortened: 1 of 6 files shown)


anthonycaccese:
thank you @danhunsaker I was working through the merge of this and your update just sorted it ❤️.  Cheers for your addition on this and thank you.

PR #33, opened 2026-06-15
Feat/plex: Autoplay Next Episode

Biduleman:
## Summary

Adds an optional "Autoplay Next Episode" setting to the Plex module. 
When enabled, whenever an episode finishes on its own the next episode of the current season will launch automatically. At the end of the season, the usual "exit to episode screen" behavior returns, as if seasons were playlists. Subtitles and audio selection will carry over based on language if they're available in the next episode. 

## AI involvement

The code was written by Claude Opus 4.8.
The testing was done manually

## Testing
Tested on both Raspberry Pi (through CRT) and MacBook (using the laptop's screen).
Tested on both regular episodes and stacked (multi-episode) files.
Verified: 
 - Natural EOF advances. User-quit (ESC/stop) does not advance.
 - The last episode in a season falls back to the detail view.
 - Transcoded episodes advance correctly.
 - Pressing "back" after several auto-advances returns to the correct episode's detail.

## Checklist

- [x] Changes work with remote-only navigation (up/down/left/right, enter, esc/backspace)
- [x] Handled sizing and positioning using the `root.sh` & `root.sw` properties (did not hardcode pixels) and kept CRT overscan in mind
- [x] Did not add tracking or analytics; only wrote settings to the local data directory if the change required storage
- [x] Follows the patterns in [ARCHITECTURE.md](https://github.com/anthonycaccese/240-MP/blob/main/ARCHITECTURE.md) and the principles in [CONTRIBUTING.md](https://github.com/anthonycaccese/240-MP/blob/main/CONTRIBUTING.md)

PR #69, opened 2026-06-23
fix: local files freeze when playback ends on its own

### LINKED ISSUE #68: [Bug]: Local Files Module - 240-mp unresponsive after completing playback of a file
### Before you file

### Area / module

Local Files module

### Platform

Raspberry Pi 4

### Display & output

CRT via composite (NTSC)

### Install method

### mpv version

240-MP v2026.06.22

### Steps to reproduce

1. Open the Local Files module
2. Select a file for playback
3. Wait for the video to finish playback or fast forward to near the end and let playback complete
4. 240-mp jumps back to the menu but freezes and is totally unresponsive to any inputs from keyboard, controller, remote, etc. I've been having issues with how my network handles ssh and haven't been able to test whether the pi is responsive to ssh commands.

### Expected vs. actual

I expect that after the video file ends to be send back to the menu but instead the app freezes and I have to reboot/unplug the pi. This does not happen if I manually back out to the menu without letting the video finish playback

### Logs / terminal output

```shell

```

anthonycaccese:
## Summary

Fixes: https://github.com/anthonycaccese/240-MP/issues/68

After "playbackFinishedNaturally" was added to the Plex module for handling auto play next episode (https://github.com/anthonycaccese/240-MP/pull/33) I missed handling it in Local Files module.  

Because of that; goBack() never ran and the app froze when playback of a video in local files ended on its own (user quit worked well and loop too so the freeze was only on natural playback end). 

So this fixes that and also refactors the three mpv exit signals (playbackFinished / playbackFinishedNaturally / playbackFailed) into one playbackEnded(pos, dur, reason)

With this a view can connect to a single handler now and can't silently drop a case (so no more freeze possibility for future cases as long as playbackEnded is used). 

This migrates Plex, Local Files, and Ambient Mode to the new logic.

FYI: @danhunsaker, @tmcleroy, @Biduleman... I apologize for missing this before but the good news is it helped harden things a bit, so hopefully this consolidated approach for playbackEnded simplifies things for any playback complete logic you may try out  in the future.  @RP2, this will need updating your Jellyfin PR likely, I'll take a look this week.

## AI involvement

Used to plan the refactor of playbackFinished / playbackFinishedNaturally / playbackFailed into playbackEnded

## Testing

Tested on RPi and MacOS for the following use cases:
- Local Files: Direct file playback (Loop off) = exits back to 240-MP correctly now
- Local Files: Direct file playback (Loop on) = works as previously, continues to loop until user exit
- Local Files: Playlist playback (Loop off) = exits back to 240-MP correctly now
- Local Files: Playlist playback (Loop on) = works as previously, continues to loop until user exit
- Plex: Episode (autoplay off) = works as previously, exits back to to item.qml
- Plex: Episode (autoplay off) = works as previously, moves to next episode until end of season or user quit
- Plex: Movie = works as previously, exits back to to item.qml
- Ambient Mode = works as previously, continues to loop until user exit

## Checklist

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -176,7 +176,12 @@ The current MPV implementation is a good reference implementation of the "browse
 1. **Launch** — `loadAndPlay(url, startSeconds, audioTrack, subTrack, ...)` starts mpv as a `QProcess`. Playback parameters are passed as mpv command-line flags: `--start=<sec>` (resume offset), `--playlist-start=<n>`, `--loop-playlist=inf`, and so on. mpv is found on `PATH` — the app never links libmpv.
 2. **Control channel** — mpv is started with `--input-ipc-server=<socket>` (a Unix domain socket at `/tmp/240mp-mpv.sock`). `MpvController` connects to it with a `QLocalSocket` and sends JSON commands via `sendCommand(QJsonArray)`. `seekTo()` and `sendKey()` (which sends mpv a `keypress` command) go over this channel — that's how the USB remote / keyboard drives mpv's OSC while it's fullscreen.
 3. **State back to QML** — `MpvController` issues `observe_property` for `time-pos`, `duration`, and `playlist-pos`, and re-publishes them as `Q_PROPERTY`s + the `positionChanged` / `durationChanged` / `playlistPosChanged` signals. A watchdog timer logs a warning if no `time-pos` event arrives for ~10 s (freeze detection).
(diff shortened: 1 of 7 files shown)


Biduleman:
Good catch I absolutely should have tested the local files, I'll make sure to remember next time.

anthonycaccese:
you're all good @Biduleman, no worries at all from my end

PR #71, opened 2026-06-24
Maintain crop setting across autoplay-next Plex episodes

tmcleroy:
## Problem

Each Plex episode launches a **fresh mpv subprocess**. The OSC **CROP** button only toggles mpv's runtime `panscan` property, which dies with the process — so when autoplay advanced to the next episode, the viewer's crop was lost and the episode started uncropped. (Local Files is unaffected: it uses a single mpv with an internal playlist, so panscan persists naturally.)

## Fix

Carry the crop across the relaunch, mirroring how the existing audio/subtitle language selection already survives an episode advance.

- **`MpvController`**: observe mpv's `panscan` property over IPC and expose it as a live `crop` QML property; add a `crop` launch parameter that emits `--panscan=<value>` so the next episode comes up already cropped (no flash of uncropped video).
- **`modules/plex/views/Player.qml`**: track the live crop via `onCropChanged` and pass it into the autoplay-advanced episode's launch.

## Behavior

- Crop set on episode 1 carries forward for as long as autoplay keeps advancing.
- Un-cropping mid-binge makes the next episode start uncropped.
- Crop does **not** leak across separate browse-and-play sessions — starting a new item from the menu creates a fresh `Player` with no crop.

## Notes

- The Pi 3 smooth-playback path can't crop at all (existing trade-off behind the Smooth Playback toggle); this applies wherever crop already works (Pi 4, Pi 5, desktop, Pi 3 with Smooth Playback off).

## Provenance & testing

AI-written by Opus 4.8 in Claude Code, then human-tested on macOS and on a Raspberry Pi 3B+ driving a CRT.

### DIFF
--- a/modules/plex/views/Player.qml
+++ b/modules/plex/views/Player.qml
@@ -44,6 +44,10 @@ FocusScope {
 
     property int lastKnownPositionMs: 0
     property int lastKnownDurationMs: 0
+    // Live mpv crop (panscan, 0..1) set via the OSC CROP button. Tracked so it can
+    // be carried into the next episode — autoplay-next spawns a fresh mpv per episode,
+    // which would otherwise reset the crop the viewer chose.
+    property real lastKnownCrop: 0
 
     focus: true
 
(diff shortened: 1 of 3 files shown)


anthonycaccese:
scope on this looks good but it got me thinking more about [this discussion](https://github.com/anthonycaccese/240-MP/discussions/31) about adding a app level setting (e.g. Auto Crop) that would allow a user define their default state for crop in one spot.

Default Off (can still be turned on as needed during playback)
On = start all video playback with panscan set to 1 (can still be turned off during playback)

and then that would flow into the work you've done in this PR so far I think where if you choose to change it during video playback - then the choice you make there persists until playback stop.  next playback start then looks at your auto crop setting again

What do you think about this as a auto-crop setting instead and handling that as the end to end feature?

tmcleroy:
@anthonycaccese thanks for making me aware of that discussion. I agree that a setting would be more appropriate and more useful for the usecase described in that discussion and in my own personal usecase where I pretty much crop everything as soon as I start playing it, auto crop would be a nice option. I'll close this PR

anthonycaccese:
@tmcleroy I just merged https://github.com/anthonycaccese/240-MP/pull/87 with the auto crop global setting.  hoping that works well for your use case.

PR #72, opened 2026-06-24
feat: add 'start on module' setting to skip module list on startup

### LINKED ISSUE #54: [Idea]: Settings - Start on Module
### Area

App shell / general

Provide an app level setting to enable the app to directly launch into a specific module.   Would still allow a user back stack to the module list but sets a default start up view in case they want always use a specific module on start up.

### Why is this useful?

For users that primarily use a single module this would give them a direct path to that module on start up (essentially saves a screen view and action)

RP2:
## Summary

Closes #54, adds an app-level setting to choose a default startup module. When set, the app skips the module list and launches directly into the chosen module. Pressing ESC/Back returns to the module list as normal. Instead of navigating from ModuleList's modulesLoaded signal (which fires during component initialization before the Loader is ready), the entry point is resolved via a C++ method and checked in the Loader's onLoaded handler after the view is fully constructed.

Changes:
- src/AppCore.h / src/AppCore.cpp - new Q_INVOKABLE startupModuleEntryPoint() resolves the saved module name to its entry point path
- Main.qml - checks the entry point in the Loader's onLoaded and auto-navigates (once per session, guarded by _startupNavigated flag)
- views/Settings.qml - new "Start on Module" list_single setting under APPLICATION section, dynamically populated from installed modules

## AI involvement

Entire implementation was AI-generated (opencode go with deepseek-v4-flash, kimi-k2.7-code, and mimo-v2.5). Human testing covered: setting the preference, restarting the app, verifying the target module loads directly, pressing ESC/Back to return to the module list, and confirming "None" preserves existing behavior.

## Testing

- Platform(s) tested: Desktop Arch Linux
- Display / output tested: HDMI desktop, DisplayPort desktop

## Checklist

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -135,6 +135,7 @@ Window {
     // --- APP-LEVEL NAV STACK ---
     property var appNavStack: []
     property var appCurrentParams: ({})
+    property bool _startupNavigated: false
 
     // --- MODULE LOADER ---
     Loader {
@@ -149,7 +150,21 @@ Window {
             }
         }
 
-        onLoaded: item.forceActiveFocus()
+        onLoaded: {
+            item.forceActiveFocus()
+            if (!root._startupNavigated) {
+                root._startupNavigated = true
+                var entryPoint = appCore.startupModuleEntryPoint()
+                if (entryPoint) {
+                    root.appNavStack.push({
+                        source: moduleLoader.source,
+                        params: root.appCurrentParams,
+                        listState: {}
+                    })
+                    moduleLoader.setSource(entryPoint, { "navParams": {} })
+                }
+            }
+        }
 
         Connections {
             target: moduleLoader.item
--- a/src/AppCore.cpp
+++ b/src/AppCore.cpp
@@ -337,3 +337,16 @@ QString AppCore::parentDirectory(const QString &path) {
 QString AppCore::homePath() {
     return QDir::homePath();
 }
+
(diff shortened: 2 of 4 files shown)


anthonycaccese:
Thank you for picking this up @RP2

In testing it I found 2 things that I thought would be helpful to add
1. Making sure the list of modules to select only includes `enabled` modules
2. storing the Id of module for the selection instead of the name (thinking long term that may help avoid conflicts on config if a module ever needs a name change).  Nice side effect of this change is that it sets a pattern for list_single to allow for display/valueToStore if we ever need it for future settings (works similarly to dynamic options that already has a pattern for display/valueToStore)

I just pushed updates to account for those.

Testing wise I think the disabled addition is working well, the guard essentially helps avoid a user getting into an odd state if they had a start on module set to a module that they then disabled (you'll see the UI switch to none in this case and during start up it will default back to the list).  The guard also helps save multiple writes to config so you'll still see the module id they had selected in this case until they make a direct change in the UI using the start on module setting itself, thats a common pattern around the app now to avoid multiple writes on settings changes so I think its all good.

Def interested in your feedback after you try it out.  If you feel good about it from there i'll move forward with the commit. 

Thank you for your work and help on it.

RP2:
I appreciate the feedback @anthonycaccese, this looks great!

Filtering to _enabled_ modules and switching to ID-based storage both make sense. The options/values pattern is a nice fit for inline settings. No issues on my end, ready to commit.

PR #75, opened 2026-06-24
feat: Support loading *multiple* custom themes

danhunsaker:
## Summary

The custom theme support is pretty nice, but does have the drawback of assuming that users will either never want to change their theme once set, or at least that they are happy to edit hex codes any time they want to switch. This change adds support for an alternate config file that supports multiple custom themes being loaded simultaneously, complete with custom names. It uses the same format, with just an extra layer to provide names as keys. It does *not* remove support for the existing single-Custom configuration.

The actual code changes are fairly simple. One new method in `AppCore` to load the themes from disk, a new codeblock in `Main.qml` to call it and add the themes to the UI, and one new codeblock in `Settings.qml` to expand the list to be able to select them. Validation in `AppCore` includes a regex to limit supported characters and lengths, in addition to the existing validation logic from the single-theme config - which has been refactored, to avoid unnecessary code duplication, into a separate (private) method.

## AI involvement

No AI used.

## Testing

- Platform(s) tested: Pi 5
- Custom Color Schemes Config Tested:
  ```json
  {
      "Abberation": {
          "primary": "#d8544f",
          "secondary": "#efa44d",
          "tertiary": "#6b6e9c",
          "surface": "#2b2d3b",
          "accent": "#8e5f95"
      },
      "Lo-Fi 90s Lighter": {
          "primary": "#f9c54e",
          "secondary": "#e76e50",
          "tertiary": "#f4a462",
          "surface": "#274754",
          "accent": "#2a9d90"
      },
      "Deliberately-Too-Long Theme Name": {
          "primary": "#66ff66",
          "secondary": "#33ff33",
          "tertiary": "#00ff00",
          "surface": "#666666",
          "accent": "#ff6666"
      },
      "\"`ILLEGAL`\" Theme Name": {
          "primary": "#ff6666",
          "secondary": "#ff3333",
          "tertiary": "#ff0000",
          "surface": "#333333",
          "accent": "#6666ff"
      }
  }
  ```

## Checklist

### DIFF
--- a/Main.qml
+++ b/Main.qml
@@ -94,7 +94,7 @@ Window {
 
         var cThemes = appCore.getCustomColorSchemes()
         if (Object.keys(cThemes).length > 0) {
-            var t = Object.assign({}, themes)
+            var t = Object.assign({}, themes, root.allThemes)
             for (var cTheme in cThemes) {
                 if (Object.keys(cThemes[cTheme]).length === 5) {
                     t[cTheme] = cThemes[cTheme]
@@ -105,7 +105,7 @@ Window {
 
         var custom = appCore.getCustomColorScheme()
         if (Object.keys(custom).length === 5) {
-            var t = Object.assign({}, themes)
+            var t = Object.assign({}, themes, root.allThemes)
             t["Custom"] = custom
             root.allThemes = t
         }
--- a/Main.qml
+++ b/Main.qml
@@ -92,6 +92,17 @@ Window {
     Component.onCompleted: {
         var cfg = appCore.get_settings()
 
+        var cThemes = appCore.getCustomColorSchemes()
+        if (Object.keys(cThemes).length > 0) {
+            var t = Object.assign({}, themes)
+            for (var cTheme in cThemes) {
+                if (Object.keys(cThemes[cTheme]).length === 5) {
+                    t[cTheme] = cThemes[cTheme]
+                }
+            }
+            root.allThemes = t
+        }
+
         var custom = appCore.getCustomColorScheme()
         if (Object.keys(custom).length === 5) {
             var t = Object.assign({}, themes)
--- a/src/AppCore.cpp
+++ b/src/AppCore.cpp
@@ -286,22 +286,10 @@ QVariant AppCore::get_installed_modules() {
     return result;
 }
 
-QVariantMap AppCore::getCustomColorScheme() const {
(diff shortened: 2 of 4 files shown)


anthonycaccese:
first, i think we should add some the color schemes from your test file as direct ones =)  ("Lo-Fi 90s Lighter" is awesome I think)

I dig this idea and there were a few people asking for this from the first overview video I had put up, so it fits requested idea too.

regex rules for name look good to me too and the validation logic refactor is appreciated.

one question I had in testing was what if a user had a custom_color_scheme.json already set and then added custom_color_schemes.json and found that with both the `custom_color_schemes` items just take on default video 1 colors as a fallback.  

I wouldn't say thats a going to be a common use case (its not likely more than a handful of people) so i think its easy enough to move forward with it as is and handle that through documentation (e.g. just update the wiki to talk about the multiple approach) then slowly migrate away from the single approach in future (add a note about deprecating it) and eventually remove it.  

But wanted to check if that might be your intention long term as well with this PR

danhunsaker:
That wasn't my intention, no - it was a bug. Turns out I inadvertently obliterated the plural-`schemes.json` values when adding the single-`scheme.json` value, by applying the static predefined `themes` without also including the updated `root.allThemes`. That has been fixed now. 😅 

I suspected a deprecation might be in order, but I'm perfectly content with pretty much any direction, moving forward.

I should ask, though. With this approach, it's possible to replace/update/change the color themes that ship with the app (try adding a `Video 1` to the `custom-color-schemes.json` ~~- it duplicates in the `Settings` list, but the `Main` colors get replaced on both~~). Is that something we're OK with ~~(aside from needing to remove the duplicate entry in `Settings`)~~, or should I try to prevent it (by skipping applying it in `Main` as well)?

anthonycaccese:
Sweeeeeet, thank you @danhunsaker

re: "With this approach, it's possible to replace/update/change the color themes that ship with the app"
I am ok with that being a possibility, would allow for a nice side effect i think to tweak internal color schemes and propose changes (no need to edit the main.qml manually for that until you land on something you like).  Like that keeps it simpler vs adding a name lookup for dedupe or a unique identifier to key off of too.

So yeah keep it as you've implemented.  Will merge now and will update the wiki today with details.

anthonycaccese:
wiki updated: https://github.com/anthonycaccese/240-MP/wiki/Customizations

PR #76, opened 2026-06-25
fix: settings - Support module-level help text

danhunsaker:
## Summary

The new settings "help" feature is currently only available on the main Settings screen. This limits the ability to assist users with more-obscure settings within module settings pages, so I went ahead and added support for module settings as well.

~~Also tweaked the display sizing a bit to support `description` text longer (or shorter) than two lines. It doesn't take wrapping into account at the moment, though, so it's recommended to keep individual help lines short.~~

## AI involvement

No AI used.

## Testing

- Platform(s) tested: Pi 5

Added a `description` value to `local_files.auto_subtitles`. This change is not included here since it isn't directly related, and makes more sense in another already-open PR or two.

## Checklist

### DIFF
--- a/views/ModuleSettings.qml
+++ b/views/ModuleSettings.qml
@@ -321,6 +321,33 @@ FocusScope {
         }
     }
 
+    // --- HELP TEXT --- (shown when a focused row has a description)
+    Rectangle {
+        id: rowHelpBackground
+        property var currentRow: moduleSettingsRoot.schemaItems[settingsList.currentIndex]
+        visible: !!(currentRow && currentRow.description)
+        color: root.accentColor
+        anchors.bottom: parent.bottom
+        anchors.left: parent.left
+        anchors.bottomMargin: root.sh * 0.1583333 //76
+        anchors.leftMargin: root.sw * 0.125 //80
+        width: root.sw * 0.75 //480
+        height: root.sh * 0.0583333 //28
+        clip: true
+        Text {
+            id: rowHelp
+            text: (rowHelpBackground.currentRow && rowHelpBackground.currentRow.description) || ""
+            color: root.surfaceColor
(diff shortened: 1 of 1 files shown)


views/ModuleSettings.qml
```diff
@@ -321,6 +321,33 @@ FocusScope {
...
+        id: rowHelpBackground
+        property var currentRow: moduleSettingsRoot.schemaItems[settingsList.currentIndex]
+        visible: !!(currentRow && currentRow.description)
+        color: root.accentColor
+        anchors.bottom: parent.bottom
+        anchors.left: parent.left
+        anchors.bottomMargin: root.sh * 0.1583333 //76
+        anchors.leftMargin: root.sw * 0.125 //80
+        width: root.sw * 0.75 //480
+        height: root.sh * (0.0291667 * ((currentRow && currentRow.description) || "").split("\n").length) //14 per line
```
anthonycaccese: would suggest keeping this to `height: root.sh * 0.0583333` which is a roughly 28px row height on 640x480

Its designed to fit within a slot in the existing row grid seen in the attached images with some slight spacing above and below it (the second image shows the slot in red)

Image

Image

That sizing accounts for a max of 2 lines height wise which I think we should try to keep to for in app help text.  and structurally would allow for wrapping when needed

anthonycaccese:
outside of the height feedback above this looks great and tested well when adding a description to a settings property in a manifest in each module.

views/ModuleSettings.qml
danhunsaker: I adjusted it because I don't know how to document the `[Forced Only]` mode without a third line, but I'll revert that now and figure the help text out later.

anthonycaccese:
thank you @danhunsaker

Issue #50, opened 2026-06-18
[Idea]: Local Files - Add support for image types


anthonycaccese:
### Area

Local Files module

### What would you like to see?

via imark7777777  on [YouTube](https://www.youtube.com/watch?v=r-gylGDoELY&lc=Ugzev1HsTKIDVcd9VG54AaABAg)
> I've been helping out with my local small town movie theater and we put a screen in the window that loops local advertisements. First one can only do photos or select videos but the photos look fine. So the second one I got one of those little boxes that can slideshow and play videos, works great even has a couple random transitions except it plays them in alphabetical order with videos first. so I can't like rename files or create a playlist to play them in a particular order because it's always gonna play the videos first. Looking at this beautiful interface and thinking well I might move to a raspberry pie if I could have some thing that would slideshow with some transition effects photos and then do videos and go through the directories in alphabetical order and not prioritizing one format versus the other. It would be very tempting. I love the color schemes well done. We are blasted with terrible user interface design granted we used to have minimalist design due to minimal capabilities but that meant you actually had to put some effort into creating a good design interface.

MPV natively handles photo assets so I'm thinking for a simple implementation i'll add the abilty for the photo module to display any image assets that exist in its media directory, set a default image display duration (and maybe add a setting with different durations).  Outside of that from my local testing so far playlist support (and even mixing images with videos in a playlist) just comes for free once the types are enabled.

### Why is this useful?

will solve an interesting and fun use case described by imark7777777 in their comment above

### Alignment with the project's approach

- [x] This fits the "browse & hand-off" approach and works with remote-only navigation (up/down/left/right, enter, esc/back).

anthonycaccese:
will be addressed by: https://github.com/anthonycaccese/240-MP/pull/95

anthonycaccese:
Done via https://github.com/anthonycaccese/240-MP/pull/95

PR #95, opened 2026-06-30
feat: local files - add support for image playback

anthonycaccese:
## Summary

adds image playback support inspired by the comment from imark7777777 in https://github.com/anthonycaccese/240-MP/issues/50

images can be opened anywhere from within the local media directory 
they can be looped, shuffled and included in playlists (and even mixed with videos in a playlist which would handle the a kiosk type use case that imark7777777 was interested in using at their local theater)

this also adds a setting to local playback to define the image playback duration (defaults to 5 seconds)

## AI involvement

used to help plan the loadAndPlay approach to make sure that image duration did not cause regressions in other modules.  that said, now that we have a few more modules I am thinking it will be a good time in the near future to re-look at loadAndPlay parameters across all again to see if there are any benefits with some refactoring.  not necessary but could be helpful.

## Testing

I tested this on MacOS and RPi on both laptop screen and CRT
I ran through each duration to make sure they matched up and tested images as standalone items and in playlists (with videos).  Looping, shuffle all worked as planned.

## Checklist

### DIFF
--- a/modules/local_files/views/Player.qml
+++ b/modules/local_files/views/Player.qml
@@ -280,8 +280,7 @@ FocusScope {
     }
 
     function isPlaylist(path) {
-        var lower = path.toLowerCase()
-        return lower.endsWith(".m3u") || lower.endsWith(".m3u8")
+        return localFilesBackend.isPlaylist(path)
     }
 
     function isImage(path) {
--- a/modules/local_files/views/Player.qml
+++ b/modules/local_files/views/Player.qml
@@ -19,6 +19,10 @@ FocusScope {
     property string resumeSetting:       "ask"
     property string subtitleMode:        "forced"
     property var    subtitleLangs:       []
+    property int    imageDurationSec:    5
+
+    // mpv subtitle-track flag derived from subtitleMode: 0 = on, -1 = forced only, -2 = off.
+    property int    subFlag:             (subtitleMode == "on") ? 0 : ((subtitleMode == "forced") ? -1 : -2)
 
     // Track last non-null values during playback for robust save on exit
     property int    lastKnownPositionMs:  0
@@ -51,7 +55,7 @@ FocusScope {
                 if (choiceIndex === 0 && startPlPos >= 0)
                     resumedFromPlaylistPos = startPlPos
                 overlayVisible = false
-                mpvController.loadAndPlay(filePath, startMs / 1000.0, 0, (subtitleMode == "on") ? 0 : ((subtitleMode == "forced") ? -1 : -2), [], subtitleLangs, loopOn, startPlPos)
+                mpvController.loadAndPlay(filePath, startMs / 1000.0, 0, subFlag, [], subtitleLangs, loopOn, startPlPos, 0.0, "", false, "", false, [], imageDurationSec)
                 event.accepted = true
             }
         } else {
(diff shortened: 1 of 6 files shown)


anthonycaccese:
@danhunsaker i thought you might be interested in checking this out with all the contributions you've made for local files, but no worries if not.

modules/local_files/views/Player.qml
```diff
@@ -19,6 +19,10 @@ FocusScope {
     property string resumeSetting:       "ask"
     property string subtitleMode:        "forced"
     property var    subtitleLangs:       []
+    property int    imageDurationSec:    5
+
+    // mpv subtitle-track flag derived from subtitleMode: 0 = on, -1 = forced only, -2 = off.
+    property int    subFlag:             (subtitleMode == "on") ? 0 : ((subtitleMode == "forced") ? -1 : -2)
```
danhunsaker, line 25: ☝🏻🎉🔥

I should have done it this way from the start. 😅🤦🏻

src/modules/local_files/LocalFilesBackend.cpp
```diff
@@ -1,13 +1,20 @@
...
-    "mp4", "mkv", "avi", "mov", "m4v", "webm", "wmv", "flv", "f4v", "mpg", "mpeg", "vob", "m3u", "m3u8"
+// Single source of truth for image types. kMediaExts (the browse filter) is built
+// from it, and QML asks isImage() rather than keeping its own copy of the list.
+static const QStringList kImageExts = {
+    "jpg", "jpeg", "png", "gif", "webp", "bmp", "tif", "tiff"
 };
+static const QStringList kMediaExts =
+    QStringList{ "mp4", "mkv", "avi", "mov", "m4v", "webm", "wmv", "flv", "f4v", "mpg", "mpeg", "vob" }
+    + kImageExts
+    + QStringList{ "m3u", "m3u8" };
```
danhunsaker: Don't we still define these two as `kPlaylistExts` or similar later on? Might be good to factor those up to here and do it once as well, while also making it more obvious why they're separate in defining `kMediaExts`.

danhunsaker (commented):
Seems pretty straightforward from here. Couple of observations, but nothing obviously wrong or broken or similar. 😅 Looks great!

modules/local_files/views/Player.qml
anthonycaccese: nah, all good!  i am more on the side of "build to get things working first" and then see "what can be refactored later".  Figure its easier to see what can be reworked when you have more to work from =)

src/modules/local_files/LocalFilesBackend.cpp
anthonycaccese: ooops yeah we do, thank you! will bring that into this

anthonycaccese: done and done

src/modules/local_files/LocalFilesBackend.cpp
```diff
@@ -164,7 +172,6 @@ QVariantList LocalFilesBackend::getItems(const QString &path) {
         return result;
     }
 
-    static const QStringList kPlaylistExts = { "m3u", "m3u8" };
     for (const QString &name : dir.entryList(QDir::Dirs | QDir::NoDotAndDotDot, QDir::Name)) {
         QString suffix = QFileInfo(name).suffix().toLower();
         if (kPlaylistExts.contains(suffix)) {
```
danhunsaker: Picking nits at this point, but `isPlaylist()` here?

anthonycaccese: great catch! I was duplicating unnecessarily for sure, added isPlaylist as the check instead.

danhunsaker:
It won't let me resolve any comment threads from here, but everything looks good on my end!

anthonycaccese:
awesome, thank you for your help!

PR #102, opened 2026-07-02
Jellyfin Module: intro/credit skip support, auth token security fix, and audio/subtitle track persistence

### LINKED ISSUE #97: [Bug]: Jellyfin Module - Audio/Subtitle track selection not persistent across show navigation
### Before you file

### Area / module

Jellyfin

### Platform

Other Raspberry Pi model

### Display & output

CRT via composite (NTSC)

### Install method

### mpv version

v0.40.0

### Steps to reproduce

1. Open a TV show episode detail in the Jellyfin module
2. Change Audio track (e.g. "English" → "Japanese") and/or Subtitle track (e.g. "Off" → "English") using the left/right arrows
3. Press PLAY, selected track plays correctly during the session
4. Navigate back to the show/episode list
5. Open a different episode (or the same episode again)
6. Different language / audio than previously selected

### Expected vs. actual

Expected (like Jellyfin browser/TV app): Once set, audio/subtitle preferences persist across all show/episode navigation and app restarts. The app remembers "I want English audio with subtitles Off" and applies that to every play regardless of which episode/show is opened.

Actual: Track selections reset to server defaults when navigating between episodes. The save_user_preferences() API call stores language preferences server-side, and applyLanguagePreferences() attempts to restore them, but the preference is only applied when the async server response arrives after the view has already rendered with defaults. Changes made during playback (via mpv's OSD) are never saved at all since save_user_preferences() is only called from Item.qml's arrow handlers. There's no local caching, so it depends entirely on the server round-trip completing before the user interacts. 

Or at least I think this is what's happening...

p.s. need to add Jellyfin to "Area / module" selection in issue form

### Logs / terminal output

```shell
What's missing / needs to change:
- Item.qml (modules/jellyfin/views/Item.qml:256-284): applyLanguagePreferences() only runs when the server prefs async response arrives after the view already rendered audioIdx = 0. The indices should be initialized from a local source (e.g. config.json under modules.com.240mp.jellyfin) so there's no flash of defaults waiting on a network round-trip
- Player.qml (modules/jellyfin/views/Player.qml:34-35): carryAudioLang/carrySubLang are in-memory only, lost on back-navigation. Should be persisted to config.json on every track change during playback so they survive view teardown
- Player.qml: No save_user_preferences() call exists, the track selections made in mpv's OSD during playback are never saved. A hook in the overlay/pause menu should persist the active track languages
- JellyfinBackend.cpp (src/modules/jellyfin/JellyfinBackend.cpp:1468-1504): save_user_preferences/load_server_preferences is server-only with no local fallback. Adding a read/write to appCore.save_setting() / appCore.get_setting() alongside the server calls would make preferences instant and offline-safe
```

### LINKED ISSUE #98: [Idea]: Jellyfin Module - Skip Intro support (intro-skipper plugin)
### Area

A new module

1. Install and configure the intro-skipper (https://github.com/intro-skipper/intro-skipper) plugin on your Jellyfin server
2. Play any TV show episode that has detected intro chapters via the Jellyfin module in 240-MP
3. Observe that no "Skip Intro" button or overlay appears during the intro sequence
4. Compare with the Jellyfin TV app which shows a "Skip Intro" button overlaid during playback

Expected: During playback, when the current position falls within a detected intro segment (as reported by the Jellyfin server API), a non-blocking "SKIP INTRO" overlay should appear (e.g. bottom-right corner). Pressing the confirm button would seek to the end of the intro segment and dismiss the overlay.

Actual: No skip-intro indicator exists. The intro-skipper plugin data (chapter markers) is available from the server but is never fetched, parsed, or presented to the user.

### Why is this useful?

Intro sequences in TV shows are repetitious and there's no reason to sit through them every episode. The server-side intro-skipper plugin already does the work of detecting intro boundaries, 240-MP just needs to read that data and give the user a one-button way to skip ahead. This is a standard feature in the official Jellyfin TV app and would make 240-MP feel significantly more polished for daily-driver use with TV series. This also allows users to skip other sections if they so desire like recaps and credits.

RP2:
## Summary

Add intro and credit skip support for the Jellyfin module using the MediaSegments API, fix a security issue where the Jellyfin auth token was leaking to mpv temp files, and add compatibility with the YouTube module's `extraArgs` parameter in `MpvController`.

**Intro / Credit Skip**

Three modes per-setting: Off, Auto (silently seeks past segments), and Button (SKIP appears in the OSD, usable with gamepad or keyboard). Settings are hidden on servers that don't have the `mediasegments` capability.

**Auth token leak fix**

The Jellyfin access token was embedded as `?api_key=<token>` in stream, subtitle, and image URLs. Mpv wrote these full URLs to `/tmp/240mp-mpv.log` and used them as keys in `/tmp/240mp-mpv-subinfo.json`, both human-readable. Now the token is passed to mpv via `--http-header-fields=Authorization:MediaBrowser Token="..."` instead (same approach Plex already uses), and `api_key` is stripped from all constructed URLs.

**YouTube module compatibility**

Rebased onto upstream/main (which includes the YouTube module). The `loadAndPlay` signature now accepts both `extraArgs` (used by YouTube for `--ytdl=yes`) and `jellyfinToken` side by side, with `--ytdl=no` gated behind extraArgs so it only disables yt-dlp for Plex/Jellyfin streams.

## AI involvement

Both features were built with the help of opencode go models (deepseek-v4-flash/pro, kimi-k2.7-code, mimo-v2.5, qwen-3.7-plus). All testing, code auditing, and research was done by me.

## Testing

Both features tested with:
- Transcode playback
- Keyboard navigation in the Player overlay
- Servers with the `mediasegments` capability

## Checklist

Closes #97
Closes #98

### DIFF
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -24,6 +24,7 @@
       "type": "list_single",
       "options_source": "dynamic",
       "options_slot": "getVideoQualities",
+      "default": "720p",
       "requires_auth": true
     },
     {
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -43,6 +43,24 @@
       "requires_auth": true
     },
     {
+      "key": "intro_skip",
+      "label": "Intro Skip",
+      "type": "list_single",
+      "options": ["Off", "Auto", "Button"],
+      "default": "Off",
+      "requires_auth": true,
+      "requires_capability": "mediasegments"
+    },
+    {
+      "key": "outro_skip",
+      "label": "Credit Skip",
+      "type": "list_single",
+      "options": ["Off", "Auto", "Button"],
+      "default": "Off",
+      "requires_auth": true,
+      "requires_capability": "mediasegments"
+    },
+    {
       "key": "logout",
       "label": "Sign out",
       "type": "action",
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -49,7 +49,7 @@
       "options": ["Off", "Auto", "Button"],
       "default": "Off",
       "requires_auth": true,
-      "requires_capability": "mediasegments"
+      "description": "Requires the Jellyfin Intro Skipper plugin on your server"
     },
     {
       "key": "outro_skip",
@@ -58,7 +58,7 @@
       "options": ["Off", "Auto", "Button"],
       "default": "Off",
       "requires_auth": true,
-      "requires_capability": "mediasegments"
+      "description": "Requires the Jellyfin Intro Skipper plugin on your server"
     },
     {
       "key": "logout",
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -49,6 +49,7 @@
       "options": ["Off", "Auto", "Button"],
       "default": "Off",
       "requires_auth": true,
+      "requires_capability": "mediasegments",
       "description": "Requires the Jellyfin Intro Skipper plugin on your server"
     },
     {
@@ -58,6 +59,7 @@
       "options": ["Off", "Auto", "Button"],
       "default": "Off",
       "requires_auth": true,
+      "requires_capability": "mediasegments",
       "description": "Requires the Jellyfin Intro Skipper plugin on your server"
     },
     {
--- a/modules/jellyfin/views/Item.qml
+++ b/modules/jellyfin/views/Item.qml
@@ -247,20 +247,12 @@ FocusScope {
 
     function _saveCurrentLangs() {
         if (!detail) return
-        var aStream = (detail.audioStreams && detail.audioStreams[audioIdx]) ? detail.audioStreams[audioIdx] : null
-        var aLang = aStream ? (aStream.language || "") : ""
-        var aTitle = aStream ? (aStream.displayTitle || aStream.title || "") : ""
-        var sLang = ""
-        var sTitle = ""
-        if (subtitleIdx !== 0) {
-            var sStream = (detail.subtitleStreams && detail.subtitleStreams[subtitleIdx - 1])
-                          ? detail.subtitleStreams[subtitleIdx - 1] : null
-            if (sStream) {
-                sLang = sStream.language || ""
-                sTitle = sStream.displayTitle || sStream.title || ""
-            }
-        }
-        jellyfinBackend.set_last_track_langs(aLang, sLang, aTitle, sTitle)
+        var aLang = (detail.audioStreams && detail.audioStreams[audioIdx])
+                    ? (detail.audioStreams[audioIdx].language || "") : ""
+        var sLang = (subtitleIdx === 0) ? ""
+                    : (detail.subtitleStreams && detail.subtitleStreams[subtitleIdx - 1])
+                      ? (detail.subtitleStreams[subtitleIdx - 1].language || "") : ""
+        jellyfinBackend.set_last_track_langs(aLang, sLang)
     }
 
     function applyLanguagePreferences(d) {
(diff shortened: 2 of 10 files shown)


RP2:
Added an additional fix that addresses audio/subtitle track persistence across items (issue #97). It stores the last-used language + displayTitle in-memory on JellyfinBackend (matching the official android-tv app's VideoQueueManager approach), so selections survive when navigating between items within a session.

anthonycaccese:
may i ask for your help with the steps to get intro-stepper set up on the server - i've installed the plugin from https://github.com/intro-skipper/intro-skipper, restarted my server and rescanned my library but I am not seeing the settings options pop up in 240-mp settings

RP2:
Just ran into this same issue, it's not related to your server setup. The plugin is working fine.

The problem was in 240-MP itself: the Intro Skip / Credit Skip settings were hidden behind a `requires_capability` check that only fires after you play a video. So they'd never show up in the settings menu fresh.

latest commit should fix it!

anthonycaccese:
my understanding reading through the code though is that "requires_capability" was implemented to make sure the settings only showed for users that had media segements set up?   was that not your intention?

Also testing after the new commit i see the settings now but am still not seeing the options to skip show up in practice.  May i ask you to change this PR to draft state?  asking because it feels lik you are currently working through things (which is all good) - just doesn't seem ready yet for an initial review.

RP2:
Yeah sorry about that, changed PR to draft.

So that was my initial intention but since it only checks `requires_capability` when a video is played that implementation wouldn't work in this use case. I'll look for another approach so that those without the plugin installed don't see those settings.

As for the button not showing up I'm not sure where that's coming from, it works on my Pi 5 and Arch desktop. The button normally lives in the default mpv playback menu we have setup for 240-MP, it's the first entry before audio during a skippable section. iirc intro skipper is separate analysis from regular library scan, try manually running "Detect and Analyze Media Segments" under "Scheduled Tasks" in the dashboard.

Let me know if that fixes it for you.

anthonycaccese:
all good, no worries. thank you for the tip "Detect and Analyze Media Segments" i'll try that in the morning.

and re: "I'll look for another approach so that those without the plugin installed don't see those settings"
thinking it can be ok to display those without the check.  especially if its not a native thing and I think your help text approach would be enough to call out that it requires work for the user to do something on their server.  

main things with that approach would be:
- diligently documenting what a user would need to do on their server (and making it clear that 240-MP isn't responsible for that set up)
- making sure there are guards in place for users that decide to turn on those settings without having the server set up stuff in place

RP2:
Sorry about all the commits, I was chasing down a weird ARM-specific crash on the Pi that I couldn't reproduce on desktop.

The "skip" settings entries are gated behind a capability probe now. They should only show up if the intro skipper plugin (MediaSegments API) is actually installed.

Agreed, documenting the plugin setup is worth adding.

anthonycaccese:
thank you @RP2, i'll try to get time to review changes this week.  Taking some time with family today after work yesterday and next up for review I'll be digging into the NFC reader changes.  

Good news is my server finished updating media segments so I got to test this for real yesterday and from a user perspective it worked really nice in practice!  both auto and button settings were smooth.

the use cases to check...
- If I don't have the plugin installed and no media segments stored
- If I have the plugin installed and no media segments stored
- If I have the plugin installed with media segments (tested and confirmed yesterday)
- If I don't have the plugin installed but do have media segments stored
- If I have media segments and the settings turned off to use them in 240-MP
- If I have media segments and the settings set to Button in 240-MP but don't interact with them in the overlay does it disappear

RP2:
Hey @anthonycaccese, good to hear!

Appreciate you thinking through those edge cases for media segments, that covers the gaps well. One question about opening the control overlay, right now I have an arrow key down press simulated to open it when a segment is detected but wanted to know if you would prefer a different approach? Like rather than simulate a button press just trigger the top level overlay opening.

Also, I recently noticed two playback session IDs in the Jellyfin logs, which made me think the direct play was attempted then failed before transcoding could begin. This could be the reason why the initial playback of videos seemed to be so slow, or at the very least wasn't helping. I don't have Plex to test against but I noticed the same pattern there so I wonder if you experience similar behavior. 

In modules/plex/manifest.json, the video_quality setting is missing a "default" field. When a user hasn't explicitly cycled the setting in the UI, nothing gets saved to config, so PlexBackend.cpp line 608 falls back to "auto" which means direct play.

Two changes needed for someone with Plex to pick up:
1. Add "default": "720p" (or any desired quality) to modules/plex/manifest.json for the video_quality setting.
2. Change PlexBackend.cpp:608 from .toString("auto") to .toString("720p").

This makes sure transcode is the default when no quality has been explicitly selected, matching the Jellyfin fix. Though, if Direct Play is the desired default, just set "auto" instead of "720p".

Verify that:
- the quality selector still cycles through all options correctly
- selecting a transcode quality (e.g. 720p) actually produces an HLS transcode stream
- and switching back to "Direct Play" still works without forcing a transcode

anthonycaccese:
@RP2 I started reviewing this and I'm sorry to ask this but could you split the different fixes/features here into distinct PRs?

Thinking one for each of the following:
- intro/credit skip support
- auth token security fix
- audio/subtitle track persistence

would like as much as possible to try to move to more focused PRs where possible just to make things more straightforward and scoped for reviews.  

Your screensaver PR is a great example of that as an approach - i'll look to update this in contributing as a principle during a future doc update.  Sorry it for not making that clearer up front, i needed to see a few different PRs to get a sense of what I preferred. 

I'll let you close this PR and do the split.  Thank you.

PR #110, opened 2026-07-05
feat(app core): screen saver for menu idle & playback pause

### LINKED ISSUE #109: [Idea]: App Core - Screen Saver
### Area

App shell / general

A screen saver that engages automatically to protect CRT displays from burn-in. I was thinking something along the lines of a bouncing module logo similar to what DVD players used to do or an image slideshow which is something Jellyfin does on the official TV app. 

Two triggers:

1. Playback pause timeout — when mpv is paused for X seconds (configurable, default maybe 60), enable the screen saver on top of the frozen video frame instead of leaving a static image burning in.

2. Menu idle timeout — when navigating the app shell (browsing libraries, settings, etc.) and no input is received for Y seconds (configurable, default 60), overlay the same ambient animation.

Both modes dismiss instantly on any keypress or gamepad input, returning to whatever was on screen before. The screen saver overlay ideally reuses the existing visuals (fonts, module logos) so nothing new needs to be built.

### Why is this useful?

240-MP targets Raspberry Pi hooked up to CRT TVs via composite video. CRTs are susceptible to phosphor burn-in when static elements (pause overlay, menu screens, frozen video frames) are left on screen for extended periods. A dedicated screensaver that kicks in automatically prevents permanent display damage without requiring the user to remember to dim or power off the display.

CRT users (like me) often pause and walk away mid-playback or leave the menu open while looking up what to watch next, and a few minutes of paused static content can leave ghosting that never goes away. This feature makes the already aging hardware last longer with zero thought required from the user.

RP2:
## Summary

Adds a screen saver to prevent CRT burn-in, controlled by a single app-level "Screen Saver" setting (OFF/30/60/120 seconds). Closes #109.

Two independent triggers share one timeout:

- **Menu idle**: new `IdleTracker` C++ class monitors global input via an app-wide event filter. After the configured timeout with no keypresses, a full-black QML overlay appears with a bouncing 240-MP logo. During mpv playback activation is blocked by the `mpvActive` property, and the timer resets on playback start/end transitions.

- **Playback pause**: `scripts/screensaver.lua` runs inside mpv via `--script=`. When the video is paused longer than the timeout, a solid black full-frame overlay with white bouncing "240-MP" text appears. The SVG logo would need conversion to raw RGBA pixels on disk (convert logo.svg -resize 144x144 logo.rgba) as a build-time step.

## AI involvement

The code was written using OpenCode Go assistance (deepseek-v4-flash and other open models). The Lua script went through several rounds of fixes with ordering, ASS overlay coordinate space (`res_x`/`res_y`), and key binding portability. All changes were reviewed and tested manually before submitting.

## Testing

- Platform(s) tested: Arch Linux (X11), Pi 5 test pending #102
- Display / output tested: 1920×1080 monitor, mpv via gpu-next VO

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ add_executable(240mp
     src/modules/youtube/YouTubeBackend.cpp
     src/player/MpvController.cpp
     src/input/InputManager.cpp
+    src/input/IdleTracker.cpp
     $<$<BOOL:${APPLE}>:src/macos_utils.mm>
 )
 
--- a/Main.qml
+++ b/Main.qml
@@ -85,7 +85,17 @@ Window {
     Connections {
         target: appCore
         function onAppSettingChanged(key, value) {
-            if (key === "color_scheme") root.currentTheme = value
+            if (key === "color_scheme") {
+                root.currentTheme = value
+            } else if (key === "screensaver_timeout") {
+                if (value === "OFF") {
+                    idleTracker.enabled = false
+                    if (screenSaverActive) screenSaverActive = false
+                } else {
+                    idleTracker.enabled = true
+                    idleTracker.threshold = parseInt(value)
+                }
+            }
         }
     }
 
@@ -117,6 +127,17 @@ Window {
         }
         root.currentTheme = savedTheme
 
+        // Screensaver: apply saved setting (C++ default is enabled=OFF).
+        if (idleTracker) {
+            var ssVal = cfg.app && cfg.app.screensaver_timeout
+            if (ssVal && ssVal !== "OFF") {
+                idleTracker.enabled = true
+                idleTracker.threshold = parseInt(ssVal)
+            } else {
+                idleTracker.enabled = false
+            }
+        }
+
         // Break declarative bindings on macOS so the C++ NSWindow override
         // in forceWindowFullScreen() isn't immediately re-fought by QML.
         if (Qt.platform.os === "osx") {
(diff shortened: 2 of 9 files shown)


anthonycaccese:
@RP2 have a bunch of updates to this to normalize things a bit further (which you can see in the latest commits)

1. Fixed a timeout bug — screensaver.lua now reads the raw script-opt via mp.get_opt("screensaver_timeout") (matching mpv-osc.lua's transcode-offset convention) instead of mp.options.read_options, which silently expected a screensaver- hyphen prefix so it was falling back to 60s always.
2. Single ownership of the setting — IdleTracker now defaults to disabled, I deleted the duplicate startup-read block in main.cpp, and Main.qml now owns both init and live changes (which is the same pattern as color_scheme). MpvController parses once with toInt() ("OFF" → 0 → disabled), closing the hole where a value below 10 left the tracker enabled or loaded the script without an opt.
3. removed flashColor from the QML bounce logo (it was never rendered or used and I think the black and white look is solid enough).
4. Lua bounce box adjusted to 400×120 to match \fs108 text, so it reverses at the frame edges instead of overshooting. That said its still just text so will need to work out the switch to the logo still.
5. Additional Cleanups — PGDWN added to dismiss keys, an unused QElapsedTimer include and "deactive" typo was fixed, I droped a redundant mpvController.position > 0 condition, and the QML overlay now handles Ctrl+Q itself (it has to call Qt.quit() directly — moduleLoader's handler is a sibling, so an unaccepted key would never reach it).
6. Updated the screensaver.lua to display the logo - the logo svg is simple so the approach for this was clean because the paths in the SVG were clean enough to convert directly to paths for drawing using ASS (no need to convert the SVG to an RGBA image)
7. fixed an mpvcontroller console logging thing - before screensaver a long pause event wasn't something I had considered so when testing this I saw that mpvcontroller was thinking a long pause might be a freeze.  I added a fix to address that so that genuine freezes are clear from just a user pause event

Also tested this on both MacOS and RPi over composite to confirm its working well on both.  When you get a chance please run through it with the latest changes just to make sure its working well on your end.  If so i'll go ahead and merge it.

anthonycaccese:
Video of a test from the RPi:

https://github.com/user-attachments/assets/de22ab5d-ca85-4b1e-84da-cdaa2fd8b20c

RP2:
Hey @anthonycaccese great work! 

I'm glad you were able to get the logo working for MPV, much better than text. Seems to be working just fine on my end so I think it's ready to merge.

anthonycaccese:
Just need to call out how awesome I think this is - have had it running on my CRT for a good part of the day and its just so fun to have in place.  Thank you RP2.

PR #114, opened 2026-07-07
Fix: jellyfin module - stop leaking auth token to mpv temp files

RP2:
## Summary

Fix a security issue where the Jellyfin auth token was leaking to mpv temp files.

**Auth token leak fix**

The Jellyfin access token was embedded as `?api_key=<token>` in stream, subtitle, and image URLs. Mpv wrote these full URLs to `/tmp/240mp-mpv.log` and used them as keys in `/tmp/240mp-mpv-subinfo.json`, both human-readable. Now the token is passed to mpv via `--http-header-fields=Authorization:MediaBrowser Token="..."` instead (same approach Plex already uses), and `api_key` is stripped from all constructed URLs.

## AI involvement

## Testing

## Checklist

### DIFF
--- a/modules/jellyfin/views/Player.qml
+++ b/modules/jellyfin/views/Player.qml
@@ -258,7 +258,7 @@ FocusScope {
             // (subTrack -1 = forced-only, a no-op when nothing soft exists).
             mpvController.loadAndPlay(streamUrl, offsetMs / 1000.0,
                                        -1, -1, [], [], false, -1, 0.0, "",
-                                       false, "", false, [], 0.0, false, jfToken)
+                                       false, "", false, [], 0.0, false, [], jfToken)
         } else {
             // Direct play: file served whole. audioIdx is 0-based → mpv's 1-based
             // --aid; subtitles come from buildSubArgs (sidecars + --sid).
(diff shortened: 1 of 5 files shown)


anthonycaccese:
Thanks for this @RP2 the header approach is a good call (and similar to what Plex does). I tested it end-to-end and pushed a few additions:

1. Fixed the token not actually reaching mpv: In a recent commit for youtube I added an extraArgs param to LoadAndPlay what wasn't accounted for yet in this PR.  jellyfinToken is the 18th parameter of loadAndPlay, after extraArgs, but both calls in Player.qml were passing 17 args. So the token was silently landing in extraArgs and the Authorization header was never sent, so with api_key stripped from URLs, playback 401'd. Simple fix just added the missing [] at both call sites.  That said, I'll keep on my radar that it may be good in the future to look at LoadAndPlay again now that we have a few modules to see if a refactor could normalize things a bit more across modules.

2. Made the mpv log/subinfo files owner-only (0600): In the spirt of your PR I thought it could be good to add a CHMOD to mpvs log file similar to what we do for Plex and Jellyfin auth files.   mpv's --log-file always logs at verbose level, which records the full command line including the header token (I verified this — it appears in exactly 2 lines). Since /tmp is world-readable on the Pi, the token was still exposed to other local accounts. Restricting the file closes that, and covers any reporting of the Plex token too.

3. I removed image_url / buildImageUrl.  Since the UI is all text and nothing calls them. rather than keep the stripped version I deleted both.

4. Made the api_key strip case-insensitive — matches the intent of the old apikey check it replaced; harmless today since Jellyfin emits lowercase, just defensive.

I verified against my live server: direct play with external SubRip subs and a forced 720p transcode both play fine with auth via header only.

PR #115, opened 2026-07-07
Feat: jellyfin intro skip

### LINKED ISSUE #98: [Idea]: Jellyfin Module - Skip Intro support (intro-skipper plugin)

RP2:
## Summary

Add intro and credit skip support for the Jellyfin module using the MediaSegments API.

**Intro / Credit Skip**

Three modes per-setting: Off, Auto (silently seeks past segments), and Button (SKIP appears in the OSD, usable with gamepad or keyboard). Settings are hidden on servers that don't have the `mediasegments` capability.

## AI involvement

## Testing

Tested with:
- Transcode playback
- Keyboard navigation in the Player overlay
- Servers with the `mediasegments` capability

## Checklist

Closes #98

### DIFF
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -43,6 +43,24 @@
       "requires_auth": true
     },
     {
+      "key": "intro_skip",
+      "label": "Intro Skip",
+      "type": "list_single",
+      "options": ["Off", "Auto", "Button"],
+      "default": "Off",
+      "requires_auth": true,
+      "requires_capability": "mediasegments"
+    },
+    {
+      "key": "outro_skip",
+      "label": "Credit Skip",
+      "type": "list_single",
+      "options": ["Off", "Auto", "Button"],
+      "default": "Off",
+      "requires_auth": true,
+      "requires_capability": "mediasegments"
+    },
+    {
       "key": "logout",
       "label": "Sign out",
       "type": "action",
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -49,7 +49,7 @@
       "options": ["Off", "Auto", "Button"],
       "default": "Off",
       "requires_auth": true,
-      "requires_capability": "mediasegments"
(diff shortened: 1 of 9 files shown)


anthonycaccese:
This looks and works great @RP2 - the probe for settings is nicely scoped to jellyfin module interactions and the interactions for button, auto all work really well.

I pushed 2 small changes, one to make sure the index of OSC is reset to 1 when Skip is present so it handles a case where you may have opened the OSC to change a setting so the index is off when skip shows up and two to align the fetchSegments probe to how its being done for probeCapabilities.

If these look and work good for you, i'll merge today

RP2:
Looks good to me! 

The focus_btn reset on prompt activation and the status >= 200 probe guard are both good additions. I believe this is ready to merge.

PR #116, opened 2026-07-07
Feat: jellyfin track persistence

### LINKED ISSUE #97: [Bug]: Jellyfin Module - Audio/Subtitle track selection not persistent across show navigation

RP2:
## Summary

Persist audio and subtitle track selection across Jellyfin items.

**Track persistence**

When playing back-to-back episodes or navigating between items, the last chosen audio language and subtitle language are remembered and automatically applied to the next item. If the same track title is available (e.g. "English [DTS-HD MA 5.1]"), it is preferred over a basic language match.

Key changes:
- Saves the current audio/subtitle language (and optionally track title) whenever the user changes tracks or starts playback
- Restores saved preferences when loading a new item, falling back to the server's language defaults if no saved preference exists
- Tracks a language-group index to disambiguate between multiple streams that share the same language
- Debounces rapid arrow-key saves to avoid redundant writes
- Fixes subTrack -2 for "subtitles off" (previously conflated with -1 for "forced only")
- Strips SubtitleStreamIndex and SubtitleMethod from the transcode URL when subs are off, preventing the server from injecting unwanted subtitle tracks

## AI involvement

## Testing

## Checklist

Closes #97

### DIFF
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -24,6 +24,7 @@
       "type": "list_single",
       "options_source": "dynamic",
       "options_slot": "getVideoQualities",
+      "default": "720p",
       "requires_auth": true
     },
     {
--- a/modules/jellyfin/manifest.json
+++ b/modules/jellyfin/manifest.json
@@ -24,7 +24,7 @@
       "type": "list_single",
       "options_source": "dynamic",
       "options_slot": "getVideoQualities",
-      "default": "720p",
+      "default": "auto",
       "requires_auth": true
     },
     {
--- a/modules/jellyfin/views/Item.qml
+++ b/modules/jellyfin/views/Item.qml
@@ -247,20 +247,12 @@ FocusScope {
 
     function _saveCurrentLangs() {
         if (!detail) return
-        var aStream = (detail.audioStreams && detail.audioStreams[audioIdx]) ? detail.audioStreams[audioIdx] : null
-        var aLang = aStream ? (aStream.language || "") : ""
-        var aTitle = aStream ? (aStream.displayTitle || aStream.title || "") : ""
-        var sLang = ""
-        var sTitle = ""
-        if (subtitleIdx !== 0) {
-            var sStream = (detail.subtitleStreams && detail.subtitleStreams[subtitleIdx - 1])
-                          ? detail.subtitleStreams[subtitleIdx - 1] : null
-            if (sStream) {
-                sLang = sStream.language || ""
-                sTitle = sStream.displayTitle || sStream.title || ""
-            }
-        }
-        jellyfinBackend.set_last_track_langs(aLang, sLang, aTitle, sTitle)
+        var aLang = (detail.audioStreams && detail.audioStreams[audioIdx])
+                    ? (detail.audioStreams[audioIdx].language || "") : ""
+        var sLang = (subtitleIdx === 0) ? ""
+                    : (detail.subtitleStreams && detail.subtitleStreams[subtitleIdx - 1])
+                      ? (detail.subtitleStreams[subtitleIdx - 1].language || "") : ""
+        jellyfinBackend.set_last_track_langs(aLang, sLang)
     }
 
     function applyLanguagePreferences(d) {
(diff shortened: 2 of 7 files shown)


anthonycaccese:
@RP2, the persistance aspects look good in testing, thank you for digging into the flow and working through that.

There were a few other apsects of the PR that I also dug into.  Specifically the the video_quality default change and a couple of the session-reporting fixes. So I ended up making a few updates while reviewing.

1) I reverted the default video quality from 720p back to auto (direct play)

I think your reasoning for making the change was spot on and the 720p default idea worked around a real problem, but it papered over a bigger cause. In my testing I found the reason "auto" wasn't reliably direct playing is that we were sending the server a whitelist of containers/codecs, and Jellyfin looks to match those by exact name. So a file with pcm_s16le audio didn't match our pcm entry, embedded webvtt subs didn't match vtt, etc. If there was any mismatch then the server silently handed back a transcode URL even with direct play set.

So because of that the code is updated for direct play to advertise a match-all profile instead of a whitelist

mpv plays basically anything, so instead of maintaining a codec list that may drift over time, the code now tells the server "this client can direct play everything." If mpv genuinely can't handle a file, then we have an existing transcode-retry fallback that will catch it and resume where you left off. Also added a log line that dumps the server's TranscodeReasons whenever direct play gets denied, so the next "why is this transcoding?" question is answerable from the logs.

2) Fixed resume points getting wiped when backing out before playback starts

The new "always report stopped" logic was reporting PositionTicks=0 when you cancelled the resume prompt (or backed out during LOADING) — and Jellyfin treats a stop at 0 as "clear the resume point." So cancelling the resume prompt was deleting the very resume point it was showing. Now the stop report falls back to the item's known offset, so the session still gets closed on the server but your place in the video survives.

3) Fixed a race that could orphan sessions on transcode retry

When direct play fails and we retry with a transcode, we report the old session stopped and immediately start a new one. The stopped reply was unconditionally clearing the current session id — so if the new session's id arrived first, it got wiped and the retried session's progress reports went out without an id (recreating the orphan-session problem this PR was fixing). The clear is now guarded so it only fires if the id hasn't changed.

I tested these updates end-to-end on macOS against a live server: fresh config direct plays (stream?static=true, dashboard shows DirectPlay), 720p still transcodes with the right caps, cancelling the resume prompt no longer touches the resume point, and no lingering sessions on the dashboard after quitting.

But please let me know if you see any issues with the above.  If not i'll merge today.

RP2:
Looks good, all solid fixes. The `match-all` profile is the right approach and the resume/session guards are both correct.

anthonycaccese:
dig it, looks like I have some merge conflicts to resolve from the other changes into main - fixing them now

RP2:
Yeah sorry about that, I tried cherry picking the commits that would result in as few conflicts as possible. In the future I'll keep work scoped to one issue at a time, this one just kinda snowballed on me.

anthonycaccese:
you're all good, happens to me too

PR #138, opened 2026-07-12
feat(nfc reader): file based mapping approach

anthonycaccese:
## Summary

The PR proposes moving from nfc_mapping.json to individual file based mapping files instead.  The idea is to make it simpler for non technical users to add mappings for their NFC cards (no need to edit a .json file)

It implements a directory for storing mappings called nfc_tags in the main config directory alongside a setting that allows a user to change that directory

And implements a file based mapping approach through .txt files.  The file shape is:
- Filename = Title
- First line of the file = CardID
- Second line of the file = Path

So a user can manage each mapping they like in individual text files with the above shape

It also adds a auto create of un-matched cards so that users won't need to manually copy IDs to create a new mapping. Just tap an unmatched card it and will create a .txt file with the ID already included (a user will just need to add their path and rename the file to give it their preferred title)

I cover the above in this video walkthrough:

https://github.com/user-attachments/assets/84927531-aef8-4149-bcc8-0acf37275c82

## AI involvement

Used to plan the approach and backend implementation for reading and writing from .txt files instead of nfc_mapping.json

## Testing

Verified on Raspberry Pi and MacOS using mapped and un-mapped cards along with cards with bad or missing paths

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,8 @@ Watch on YouTube: https://youtu.be/r-gylGDoELY
 ### NFC Reader Module ([Wiki](https://github.com/anthonycaccese/240-MP/wiki/Module:-NFC-Reader))
 - Start video playback via NFC cards
 - Tested reader support: `ACS ACR122U`
-- Uses a UID-to-video mapping file (`nfc_mapping.json`) to determine which video to play for a given card
+- Maps cards to videos via per-card text files in the `nfc_tags` data directory — the filename is the display title, line 1 the card UID, line 2 the video path or URL
+- Tapping an unknown card auto-creates a stub tag file for it; rename the file and add a path line to map the card
 
 ### Global
 - [Color Schemes](https://github.com/anthonycaccese/240-MP/wiki/Customizations)
--- a/modules/nfc_reader/manifest.json
+++ b/modules/nfc_reader/manifest.json
@@ -11,6 +11,12 @@
       "default": "OFF"
     },
     {
+      "key": "tags_directory",
+      "label": "Tags Directory",
+      "type": "directory_browser",
+      "default": ""
+    },
+    {
       "key": "resume_playback",
       "label": "Resume Playback",
       "type": "list_single",
--- a/modules/nfc_reader/views/Items.qml
+++ b/modules/nfc_reader/views/Items.qml
@@ -11,15 +11,7 @@ FocusScope {
     signal navigateTo(string path, var params, var listState)
     signal goBack()
 
-    property string errorMessage: ""
-
     focus: true
-
-    Component.onCompleted: {
-        errorMessage = nfcReaderBackend.mappingLoaded ? "" : "REQUIRED FILES NOT FOUND\n\n"
-                             + "ADD NFC_MAPPING.JSON\n\n"
-                             + "PLEASE SEE THE WIKI FOR DETAILS"
-    }
 
     Keys.onPressed: function(event) {
         if (event.key === Qt.Key_Escape || event.key === Qt.Key_Backspace || event.key === Qt.Key_Back) {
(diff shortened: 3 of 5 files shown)


anthonycaccese:
@jmcerrejon please let me know what you think of this proposal for nfc reader.  it moves from nfc_mapping.json to individual file based mapping files instead.  My idea behind this change is that I think this will make it easier for non-technical users to create mappings (they can manage .txt files instead of editing a .json file)

I added a video overview of the implementation here to review:

https://github.com/user-attachments/assets/84927531-aef8-4149-bcc8-0acf37275c82

jmcerrejon:
For non-technical users, it is easier, that's true, but take into account the performance,...What is better?

I thought JSON because you know, we are developers😉 and handling a dozen files for read operations is very intensive for I/O even CPU, and we are giving support to Raspberry Pi 😅

Once I said that, check (or ask AI) if it's OK with the new way using txt vs JSON, or maybe we can use a more friendly format in a file (YAML?). If performance is the same or does not change a lot, txt is a good format. Go for it 😉

Another question: What if we add a new feature for two or more paths for a card in the future? This is a possible scenario maybe easier to solve with JSON or even JAML.

anthonycaccese:
Thank you @jmcerrejon.  I was thinking a bunch about performance during the scoping and testing of this.

I def agree JSON is the faster format to load, but the good news is the implementation tries to mitigate for this via `scanTagsDir()`: it reads the directory once and builds an in-memory QHash of UID → {path, title}. 

So in regular use a mapped card tap will never touch the filesystem, it's a pure hash lookup (which is likely microseconds regardless of file count). That scan runs at startup and re-runs only when an unknown card is tapped (which is also how a newly added tag file gets picked up without restarting the app, and it's rate-limited so a card sitting on the reader won't trigger repeated scans).

For the scan itself, the files are two lines each, so the cost will more be syscall/metadata overhead rather than data. Rough math for an extreme case of 500 mapped cards: macOS SSD ≈ single-digit ms; RPi4 SD warm cache ≈ low tens of ms; RPi4 SD cold cache (first scan after boot) is the only case that could be noticeable, maybe a few hundred ms and that startup scan warms the cache for everything after it anyway.

At what I am assuming will be realistic library sizes (I tested ~100 cards) its super quick, and if we ever see users with thousands of cards we have a few things to try (stub files are already named by UID, so a miss lookup could become a single stat instead of a re-scan as an idea).

Re YAML: I love YAML but I think it would reintroduce similar fragility for non-technical users (indentation/syntax to break), plus Qt has no built-in YAML parser so it'd add a dependency on both platforms. My thinking for .txt is that bare text lines have no syntax to get wrong 😄

Re multiple paths per card: can you say more about the scenario you're thinking through? I think the .txt format could extend to it naturally... the title is the filename, line 1 is the UID, and lines 2+ can be paths; today the logic just stops reading after line 2, so it's a one-line change when we want it. So I don't think txt closes that door if we want to open it later.

jmcerrejon:
Great news! So txt is our file format 😊

Do you remember when you bought a blank VHS tape and recorded two or more movies/series/TV shows in the past?

The idea is to add to a card more than only a video file. For example, I’ll have a card and if I put it on the NFC reader, a menu shows me: Terminator I and Terminator II (In my universe the rest doesn't exist)😂
It’s like a playlist for each card. Do you understand me?

anthonycaccese:
Ah yes I understand, so we could kind of create a overlay that shows the items for selection and then on select start playback (or present a resume question if there is already watch history on that item).  yeah pattern wise that makes sense and i think would be pretty easy to support with our current patterns.  

I'll merge this for now and then start working through the merge into main after that (going to do an -rc build to test it end to end)

PR #146, opened 2026-07-14
fix(mpv): pi5 hdmi out aspect ratio

anthonycaccese:
## Summary

the pi5 can sometimes inconsistently report its resolution for CRT output as 704×432 which can result is a horizontally squished video during playback.  we tried to solve for this directly in mpvcontroller via a  hardcoded monitorpixelaspect property but that ended up having 2 downstream impacts... 1) if the pi5 reports things correctly we've now overcompensated for this and 2) on hdmi out for the pi5 it also gets applied which causes 16:9 output to be stretched.  in both cases with the approach we previously took you could only change this through mpv_video_args in our config.json.

so thinking about this and using it practice for the last few weeks I think its more flexible to remove this hardcode and use the native native options for solving this directly instead (e.g. either mpv.conf or cmdline/configtxt at the OS level).  Its not consistent enough to be managed as a hardcoded 240-MP concern from my testing on the pi5 in both composite and hdmi out use cases.

This also fixes a minor mistype in the HDMI config instructions for Pi5

## AI involvement

None 

## Testing

Traced this on a Pi5 output to both a CRT and HDMI and found the inconsistency that can occur with CRT output can be also compensated for sometimes via config.txt or cmdline.txt or through mpv.conf.  

## Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -161,7 +161,7 @@ However, if you already have Raspberry Pi OS set up and working on your TV then 
     [pi5]
 
     # Drivers & Video
-    dtoverlay=vc4-kms-v3d,cma-512,composite=1
+    dtoverlay=vc4-kms-v3d
     
     # --- Global ---
     [all]
--- a/src/player/MpvController.cpp
+++ b/src/player/MpvController.cpp
@@ -669,11 +669,12 @@ void MpvController::appendVideoArgs(QStringList &args) const {
             else
                 args << "--vo=drm" << "--hwdec=v4l2m2m-copy";
         } else {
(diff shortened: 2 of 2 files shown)


anthonycaccese:
@RP2 as a heads up on this, i've updated the wiki here for the Pi5 with a note of the way to compensate for this via mpv.conf if it occurs during output to a CRT from a pi5.  in testing the pi5 more thoroughly over the last few weeks I found this not to be consistent enough to hardcode directly into our args (our hardcoded args can only be modified by using mpv_video_args which also means users need to copy in hwdec and vo settings when they do so)

Wiki details: https://github.com/anthonycaccese/240-MP/wiki/Hardware-Testing#crt-tv-output-3

PR #148, opened 2026-07-14
Add independent X/Y crop overscan controls

iurisouz4:
## Summary
- Add separate app settings for CRT overscan compensation:
  - Crop Overscan X (horizontal)
  - Crop Overscan Y (vertical)
- Expand overscan adjustment range to 0%..30% (2% steps).
- Apply axis-specific mpv scaling (`--video-scale-x`, `--video-scale-y`) when Auto Crop is enabled, so horizontal/vertical fit can be tuned independently.
- Keep backward compatibility with existing `crop_overscan` config by using it as fallback when new axis keys are not set.
- Document the new CRT tuning flow in BUILDING.md.

## Why
On CRT setups, overscan and geometry often require independent horizontal and vertical compensation. A single global overscan value cannot reliably fit both axes.

## Manual testing
- Verified settings are visible in app settings with 0%..30% options.
- Verified values persist through config and are read at playback start.
- Verified axis-specific scaling arguments are applied only when Auto Crop is enabled.
- Verified legacy `crop_overscan` still works as fallback.

## Media
imageimageimage

## AI use disclosure
AI assistance was used to help draft and implement this change (code edits and PR draft text). The final changes were reviewed and validated before submission

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -242,6 +242,15 @@ This config is read at each playback event, so a change applies on the next play
 
 The trade-off with this approach: the copy path didn't look like it could reliabilty play back 1080p on the Pi 3 in my testing. I found it can easily peg the CPU and cause stuttering. So enabling crop on a Pi 3 means keeping your source content to **720p and below**. Ultimately its your call: smooth 1080p without crop (keep the default), or enable crop with a 720p ceiling using --hwdec=v4l2m2m-copy.
 
+### CRT overscan tuning (Auto Crop)
+
+If **Auto Crop** fills the screen but goes a bit past the visible CRT area, use:
+
+- **Settings → Crop Overscan X** (horizontal fit)
+- **Settings → Crop Overscan Y** (vertical fit)
+
+Both apply only when Auto Crop is ON and can be tuned independently (0% to 30%).
+
 ## Debugging & logs
 
 240-MP logs to **stdout/stderr** via Qt's `qDebug` / `qWarning` (used throughout `AppCore`, `MpvController`, and the module backends). The trick is knowing where that output goes depending on how you launched the app.
--- a/src/player/MpvController.cpp
+++ b/src/player/MpvController.cpp
@@ -9,6 +9,7 @@
 #include <QDateTime>
 #include <QRegularExpression>
 #include <QDebug>
+#include <algorithm>
 
 #ifdef Q_OS_LINUX
 #include <fcntl.h>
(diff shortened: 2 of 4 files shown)


RP2:
This would be so much nicer than going into the TV's service menu!

iurisouz4:
My TV doesn't even have that option 😥

anthonycaccese:
Awesome @iurisouz4 thank you for this and a great use case.  I'll make time to review soon.

anthonycaccese:
hi @iurisouz4, i've had time to use this over the last few nights, thank you for a clean scope and easy to follow comments.

That said, in using it got me wondering about instead using the `-panscan` steps settings in place of `--video-scale-x`/`--video-scale-y`

I understand that `-panscan` would not enable the fine tuning of the width and height that `--video-scale-x`/`--video-scale-y` provides.  but x/y feels a bit more like a solve for geometry (that I would prefer to keep to TV settings or mpv.conf).   `-panscan` feels like it could be clearer to use because it would enable control over how much zoom auto-crop sets while also keeping aspect ratio intact.

Essentially what this made me think of is adding a new setting called "Auto Crop Initial Zoom" and then provide the steps from 10% to 100% to allow a user to set their default starting crop for panscan when auto crop is on.  In addition... update the crop button on the OSC menu to allow it to step through the same range when its pressed (the same as pressing `w` does when MPV is running on its own)

When you have set `--video-scale-x`/`--video-scale-y` in your testing... do you find that you have been able to set it and forget it? (e.g. the same settings for all video playback)  If yes have you tried setting it directly in mpv.conf to see if that would work for your set up?

iurisouz4:
Hi @anthonycaccese, initially, I tried to adjust the zoom using panscan. However, in my setup (a Raspberry Pi 3B+ with an HDMI-to-Composite converter), even 4:3 videos still had horizontal black bars. That's why I chose to adjust both axes directly. Now it works flawlessly; I only had to set it once and everything looks perfect, even 16:9 videos. By the way, I haven't tested setting this directly in mpv.conf yet, but it's something worth checking out.

anthonycaccese:
thank you @iurisouz4, when you have a moment to try setting those values in mpv.conf please let me know, I ran a test on my end and it worked well for me but it would be helpful to confirm.

I understand your rationale however the more that i've played with this I am thinking that it may make sense to set it and forget it in mpv.conf instead of adding X/Y adjustments directly into 240-MP (considering it may be adapter/capable and tv dependent)

RP2:
The biggest value with this being within the app itself is the ability to easily preview changes. It's true that if it's set and forget it might be a little bloated to have a constant settings entry within the app forever... but I still think there's value in avoiding the need to SSHing and adjusting the config in a text editor multiple times to dial in the proper crop. This might be where adding a greeter/setup wizard might be a nice addition. This way all the one time setup things live there and don't crowd the normal menus. For example the `monitorpixelaspect=0.82` setting.

anthonycaccese:
> The biggest value with this being within the app itself is the ability to easily preview changes 

I agree with that sentiment @RP2, the main thing that gives me pause about the X/Y approach is that can allow a user to easily get into an odd aspect ratio setup.   That's why I was proposing the idea of approaching it like this instead:

> That said, in using it got me wondering about instead using the -panscan steps settings in place of --video-scale-x/--video-scale-y
I understand that -panscan would not enable the fine tuning of the width and height that --video-scale-x/--video-scale-y provides. but x/y feels a bit more like a solve for geometry (that I would prefer to keep to TV settings or mpv.conf). -panscan feels like it could be clearer to use because it would enable control over how much zoom auto-crop sets while also keeping aspect ratio intact.

What are your thoughts on that?

RP2:
I think this might be where the use of a test pattern might help since I know when I was messing around in the TV service menu (to fix the initial crop) the fix for width was a different value than height. Unfortunately that's the nature of analog signals and displays, the best we can do is try to guide the user in the right direction. 

Though I do think your approach makes sense too, maybe an "advance" tuning gate/toggle for the independent x/y scale as an option for those who have a test pattern pulled up. This way users can get 90% of the way there with the video aspect ratio locked, but that last 10% they can optionally tweak themselves if their TV is like mine and needs some re-calibration.

anthonycaccese:
going to close this PR based on our discussion so far, thank you iurisouz4 for opening it and the time you spent working on it.  I'll look to explore adding an auto-crop "steps" setting but for aspect ratio tuning for horizontal/vertical please leverage mpv.conf settings directly.  that will accomplish what you were aiming for and will allow you to tune it even further for your direct set up.

PR #163, opened 2026-07-15
feat(input): remote/keyboard button remapping

plyrthn:
## Summary

Adds a "Remote Controls" screen under Settings for remapping an extra physical
button to one of the six navigation actions (up/down/left/right/select/back).
This is additive: it never removes an action's existing default key, it only
adds one more key that also fires that action, so a bad remap can't lock
someone out of the menus.

Covers three input paths, since a lot of cheap USB remotes split themselves
across multiple HID interfaces:
1. Ordinary keyboard keys, via Qt's normal QKeyEvent delivery.
2. "Consumer Control" HID buttons (Home, Back, Menu, colored buttons, zoom,
   etc.) — these show up as a *separate* /dev/input/eventN device from the
   keyboard interface and never reach Qt as key events at all, so this reads
   that device directly and feeds it into the same remap table.
3. Mouse button clicks — some remotes wire their center/OK button as a literal
   left click instead of any kind of key.

Also fixes a real bug in `AppCore::get_setting()`: it wasn't symmetric with
`save_setting()` for dot-notation keys (e.g. `remote_keymap.select`) —
`save_setting` correctly nested those under a sub-object, but `get_setting`
only ever looked for a literal flat key with a dot in it, which never existed,
so any dotted setting silently always read back as invalid. This affects any
future dotted setting, not just this feature.

Files: `src/input/InputManager.h/.cpp`, `views/RemapControls.qml` (new),
`views/Settings.qml`, `src/AppCore.cpp`, `src/main.cpp`

## Testing

Tested with a physical USB remote exposing Keyboard/Consumer Control/Mouse
interfaces. Verified capture works for all three input types, verified
default bindings still work after remapping, verified reset-to-defaults,
verified a held key's OS autorepeat doesn't self-complete a capture with the
wrong key. Not tested on macOS.

## Checklist

### DIFF
--- a/assets/images/keyboard.svg
+++ b/assets/images/keyboard.svg
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg width="100%" height="100%" viewBox="0 0 22 18" version="1.1" xmlns="http://www.w3.org/2000/svg" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2;">
+  <path d="M 0 18 L 0 0 L 22 0 L 22 18 L 0 18 Z M 20 2 L 2 2 L 2 16 L 20 16 L 20 2 Z M 10 6 L 8 6 L 8 4 L 10 4 L 10 6 Z M 6 6 L 4 6 L 4 4 L 6 4 L 6 6 Z M 17 14 L 5 14 L 5 12 L 17 12 L 17 14 Z M 12 10 L 10 10 L 10 8 L 12 8 L 12 10 Z M 16 10 L 14 10 L 14 8 L 16 8 L 16 10 Z M 14 6 L 12 6 L 12 4 L 14 4 L 14 6 Z M 18 6 L 16 6 L 16 4 L 18 4 L 18 6 Z M 8 10 L 6 10 L 6 8 L 8 8 L 8 10 Z" style="fill:white;"/>
+</svg>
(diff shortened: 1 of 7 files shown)


anthonycaccese:
Thank you @plyrthn, I was unsure about adding an in-app tool for this because they can get complex but I think the approach you took is nicely scoped and a good first approach for this (very much appreciate the choice to make it additive to defaults vs removing defaults)

The main gap I ran into during testing was that InputManager didn't know when a capture is in progress, which for me was causing three related bugs:

1. Rebinding an already-bound key captured the wrong key. The app-level eventFilter consumes any key already in the remap table and posts the action's default key before the overlay sees it, so rebinding key X from action A to action B saves Qt::Key_Return (or whatever A's default is), not X.
2. Capturing a mouse or consumer button fired the just-bound action on the same press. finishCapture → save_setting → appSettingChanged → loadKeyRemap all run synchronously inside the emitting handler, so the remap lookup right after the auxButtonPressed emit now hits and delivers the action. Bind Select this way and the stray Return re-opens the capture overlay.
3. Gamepad presses during capture bound a bogus literal Qt::Key_Return, the overlay can't distinguish synthesized events from real keys.

So the latest commit proposes an approach to address all three using an explicit capture mode (setRemapCapture) that suspends remap delivery and routes all three input paths through auxButtonPressed while the overlay is up, plus a few smaller fixes enumerated in the commit message (held-direction auto-repeat, double-click handling, one-button-per-action and display/language tweaks).

Please let me know what you think after review and testing, I only have a single remote to test with so i'm not sure I was able to give it as robust a test as you may be able to.  Thank you again.

plyrthn:
Let me check!

anthonycaccese:
thank you, no rush at all

plyrthn:
Works great!!

anthonycaccese:
great, thank you, I'll aim to work through your other PRs over the next week

plyrthn:
WOOT! Appreciate you brother.

PR #164, opened 2026-07-15
fix(mpv): ambient mode audio device selection

plyrthn:
## Summary

Ambient mode's companion audio track (and any unmuted mpv audio) could go
silent because ALSA's "default" pcm device resolves to whatever card
enumerates first — on my Pi 5 that was a USB remote dongle's stub audio
interface, not the real HDMI output, and that path dead-ends in a broken
dmix/pipewire fallback ("couldn't open play stream").

Adds runtime detection of the correct `vc4-hdmi` ALSA card (parsing
`/proc/asound/cards`, preferring whichever HDMI connector `/sys/class/drm`
reports as connected) and passes it explicitly via `--audio-device`, instead
of leaving mpv on "default". Falls back to leaving mpv's audio-device alone
if no vc4-hdmi card is found, so this is a no-op on macOS or non-Pi hardware.

Extracted the detection into `src/linux_audio_utils.h/.cpp` since both
`MpvController` and `AmbientModeBackend` need it — mirrors the existing
`src/macos_utils.h/.mm` pattern for a shared platform helper.

Also adds a bounded auto-restart watchdog for the ambient audio process: it
has no video to fall back on, so if it dies on its own (an ALSA hiccup, a
transient USB/device stall) the screen kept playing silently with nothing to
notice or recover. Mirrors `NfcReaderBackend`'s existing wedge-detection/
respawn pattern for the same class of failure.

Files: `src/linux_audio_utils.h/.cpp` (new), `src/player/MpvController.cpp`,
`src/modules/ambient_mode/AmbientModeBackend.h/.cpp`, `CMakeLists.txt`

## Testing

- Display / output tested: CRT via S-Video, audio via HDMI-to-composite converter

Confirmed audio plays through the correct output. Confirmed the watchdog
restarts the audio process after killing it by hand mid-session. Not tested
on macOS (the code path is Linux-only and a no-op elsewhere).

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,7 @@ add_executable(240mp
     src/input/IdleTracker.cpp
     src/update/UpdateManager.cpp
     $<$<BOOL:${APPLE}>:src/macos_utils.mm>
+    $<$<NOT:$<BOOL:${APPLE}>>:src/linux_audio_utils.cpp>
 )
 
 target_compile_definitions(240mp PRIVATE APP_VERSION="${APP_VERSION}")
--- a/src/linux_audio_utils.cpp
+++ b/src/linux_audio_utils.cpp
@@ -1,0 +1,41 @@
+#include "linux_audio_utils.h"
+#include <QFile>
+#include <QDir>
+#include <QRegularExpression>
+
+// ALSA's "default" pcm can resolve to card 0, and on this hardware card 0 is
+// whatever happened to enumerate first, including a USB dongle (e.g. the
+// remote receiver) that exposes a stub audio interface with no real playback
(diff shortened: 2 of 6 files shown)


anthonycaccese:
@plyrthn thank you, the approach you took for this makes sense but in using it I opted to add https://github.com/anthonycaccese/240-MP/pull/209 as a way to handle this vs bringing this in directly.  I've run into so many variances with audio output on the Pi that my thinking is this can be ideally handled at either the OS level (through alsa config) or via the prop that MPV offers to be set via mpv.conf.  So am going to close this PR without a merge with that in mind.  That said I am glad to discuss further.

plyrthn:
Sounds good!!

PR #165, opened 2026-07-15
feat(ambient mode): shuffle & auto-launch settings

### LINKED ISSUE #154: [Idea]: Add Shuffle Playback Option to Ambient Mode Module
### Area

Ambient Mode module

Allow music playlists to be shuffled when used as the audio track in Ambient Mode

### Why is this useful?

Comment from https://github.com/anthonycaccese/240-MP/discussions/129#discussioncomment-17589137
> I like the idea of having a music playlist with an ambient video loop, but I think having the same track order each time I turn it on could get old. An option to shuffle the playlist each time I turn on the Ambient mode could keep things fresh!

plyrthn:
## Summary

Adds three ambient mode settings: Shuffle Video Playback, Shuffle Audio
Playback, and Auto-Launch Playback (skip the picker screen and start playback
immediately — useful when Ambient Mode is set as the app's startup module).

Each shuffle toggle is independent: auto-launch only randomizes the video
pick when Shuffle Video is on, and only randomizes the audio pick when
Shuffle Audio is on; otherwise it falls back to the last-viewed video / no
companion track, same as manually pressing START would.

Fixes a QML timing race where auto-launch could silently fail: the
`navigateTo` signal could fire from `Component.onCompleted` before the
module's `Connections` block had finished rebinding to the freshly-loaded
item one tick later, so the signal went out before anything was listening.
Deferred via `Qt.callLater()`.

Files: `modules/ambient_mode/manifest.json`,
`modules/ambient_mode/views/Items.qml`, `modules/ambient_mode/views/Player.qml`

## Testing

Verified auto-launch fires reliably across multiple reboots with Ambient Mode
set as the startup module. Verified each shuffle toggle independently.
Verified the picker screen (START button) still works unchanged when
auto-launch is off. Not tested on macOS.

## Checklist


anthonycaccese:
Just a note to say i've been heads down working on a weather module so have not spent the time i originally planned to with this yet.  I plan to wrap my changes for weather by the weekend and will look to shift focus to this (and other open PRs) then.

plyrthn:
No worries brother!!!

anthonycaccese:
@plyrthn playing with this today and had a few questions:

- May I ask about use case ideas for the autostart setting?  I see it only fires when "Start on Module" is set to Ambient Mode + Auto Launch Playback is set to on in the module settings.  What are some of the use cases where that combo might work for your use cases? I hadn't considered the idea of booting into Ambient Mode and having it already playing something back, I think its cool but just wanted to hear when that might come up
- For the shuffle options, i'm wondering if these could work as inline selection options instead of settings options so that they allow you to shuffle even without autostart on.  essentially a "shuffle" option appended to the end of the list of available files in each setting.   What do you think of that as an idea?

plyrthn:
Hey! So autostart was the idea I primarily wanted this for a nice background video/audio while I work in my office! So I turn on my switch connecting my crt and my pi and it autoboots into random audio and video! So it's just for the ambiance without having to set it up every time! 

I like that!! I would argue putting it BEFORE (or looping around) so "video audio" is first then hitting left would make it shuffle? :) I cannot remember if it loops around today?

anthonycaccese:
> I primarily wanted this for a nice background video/audio while I work in my office! So I turn on my switch connecting my crt and my pi and it autoboots into random audio and video

sweet, I dig it =)

> I cannot remember if it loops around today?

Video does / Audio doesn't but it can =)  I'll work on that next and add it as a commit to this PR (will remove those options from settings as part of it and also make sure that when auto start is enabled it defaults to shuffle still (I think thats a clear enough use case where it should auto shuffle like you mentioned)

plyrthn:
Awesome!!! :)

anthonycaccese:
ok sorted @plyrthn  (at least from my testing so far).  

video preview: 

https://github.com/user-attachments/assets/7fa91ab2-0d6a-4762-a3d7-db908fb008f8

I've done a regression test against the rest of the module after merging in main to check things so I feel pretty good about those but I would appreciate a test from your end on the ambient mode changes one more time just to make sure it still works as you were aiming for

AchillesPDX:
Hey @anthonycaccese just an FYI: for testing, after you quit 240-MP you can press up on your remote and then OK on the remote and it'll launch 240-MP again without needing to use the keyboard.

PR #166, opened 2026-07-15
fix(app core): CJK glyph fallback for titles

plyrthn:
## Summary

Japanese/Chinese/Korean file and song titles rendered as tofu boxes — the
app's only bundled font (VCR OSD Mono) is Latin-only, and there was no other
font on the system to fall back to.

Bundles `assets/fonts/unifont.otf` (GNU Unifont, dual OFL/GPL+font-exception
licensed, ~5MB) and loads it via a second `FontLoader` in Main.qml. It's a
blocky bitmap-style font so it actually matches the retro CRT look instead of
a mismatched serif/sans fallback.

I initially tried Qt's `font.families` fallback-list property, but this
build's Qt (6.8.2) doesn't expose a `families` property on the QML font value
type at all (confirmed via its `plugins.qmltypes`, only singular `family`
exists) — assigning it is a hard QML error, not a soft warning, and broke the
whole app on startup. Reverted that. The actual fix needs no changes to any
view file at all: registering Unifont via `FontLoader` adds it to Qt's
fontconfig-backed font database, and Qt's own missing-glyph fallback picks it
up automatically for every existing `font.family: root.globalFont` binding.

Files: `Main.qml`, `assets/fonts/unifont.otf` (new)

## Testing

Confirmed Japanese and Chinese titles render as real characters. Confirmed
Latin titles are unaffected (still VCR OSD Mono). Not tested on macOS.

## Checklist


anthonycaccese:
Thank you for this fix, good simple fallback set up.  I added credit and license details for unifont in a follow up commit but overall looks good, let me know if you see any issues with my credit text

plyrthn:
Can't believe I forgot that! Thank you! Looks good!

PR #174, opened 2026-07-18
feat(app core, jellyfin, local files, youtube): add list wraparound

ToTheXtreme64:
## Summary

Added list wraparound to all menus I could find related to the app core, settings, jellyfin, local files, and youtube. So when you press up at the top of a list it brings you to the bottom and when you press down at the bottom of a list it brings you to the top. 
## AI involvement

Not used at all

## Testing

I tested on MacOS with my Jellyfin server, some test files, and the example subscription and playlist files. 

## Checklist

- [x] Changes work with remote-only navigation (up/down/left/right, enter, esc/backspace)
- [x] Handled sizing and positioning using the `root.sh` & `root.sw` properties (did not hardcode pixels) and kept CRT overscan in mind
- [x] Did not add tracking or analytics; only wrote settings to the local data directory if the change required storage
- [x] Follows the patterns in [ARCHITECTURE.md](https://github.com/anthonycaccese/240-MP/blob/main/ARCHITECTURE.md) and the principles in [CONTRIBUTING.md](https://github.com/anthonycaccese/240-MP/blob/main/CONTRIBUTING.md)

Definitely needs some more testing, but seems to be working great with what I have tested. I am unable to work on the other modules so those were left alone for now. The other pull request I opened related to the letter nav on collections will definitely need some tweaks to work with this pull request. A much needed QOL change.

Issue #153, opened 2026-07-15
[Idea]: Emby Module


freemchr:
### Area

A new module

### What would you like to see?

Hi Anthony, thanks for creating 240-MP. I run Emby instead of Jellyfin. Since Emby and Jellyfin have similar APIs, would Emby support be feasible, or is there already an unofficial way to use Emby with 240-MP? I'd be happy to test if you're interested.

### Why is this useful?

Support for Emby

### Alignment with the project's approach

- [x] This fits the "browse & hand-off" approach and works with remote-only navigation (up/down/left/right, enter, esc/back).

anthonycaccese:
Thank you @freemchr, taking a quick view of their docs here: https://dev.emby.media/doc/restapi/index.html I think a module would be possible.  But I don't have any experience with using Emby directly (my main server set up is Plex).  

It would be fun to explore for sure but before doing that i'd want to set up a server to play with it and see how it works day to day (plus try out their clients).  To try to set expectations it will be a while before I have the time do that though.  I'll keep this open though in case someone else would be interested in picking it up sooner. 

Some initial questions:
- Is Emby mostly free to use? (set up a server and run localy?)
- Would you know if accounts are tied to a server or if there is a centralized account system like Plex (wanting to see if its more like Jellyfin or Plex account wise)
- From your experience with it have you found that most users you've talked with with Emby setups have one server or many?

freemchr:
Hi Anthony

Thanks for the reply. I’m happy to try and built it as I have some experience in building apps. Could try and clone the repository and built the module if OK with you. 

In response to your questions

Is Emby mostly free to use? (set up a server and run localy?)

Yes it’s free but they have paid version too. 

Would you know if accounts are tied to a server or if there is a centralized account system like Plex (wanting to see if its more like Jellyfin or Plex account wise)
To the server that is run usually on a docker. It’s similar to jellyfin. Nearly the same. 

From your experience with it have you found that most users you've talked with with Emby setups have one server or many?

Usually one server not many. 

Let me know if you are OK with me having a go at it.

anthonycaccese:
thank you, yes i would def be ok with you having a go at it.  Since it may be close to Jellyfin take a look a the approach @RP2 took when building that module in `/src/modules/jellyfin` (c++ source) and `/modules/jellyfin` (manifest, qml and assets) - that could be a good base to work from.  And I'll def help with any questions as you play with things.

freemchr:
[...]

PR #183, opened 2026-07-23
Add Emby module

freemchr:
## Summary

Adds a self-contained **Emby** module (`modules/emby/` plus `src/modules/emby/`, one `registerModule` call), built on the Jellyfin module as suggested in #153 and adapted for Emby's differences. Ships disabled by default.

- **Auth** (genuine Emby work, not ported): username/password (`/Users/AuthenticateByName`) plus **Emby Connect**, a cloud sign-in that discovers the account's servers and exchanges the per-server access key for a local token, with a picker when there's more than one server.
- **Browse/playback**: libraries, shows/seasons, collections, Continue Watching / Next Up, track selection, direct play with transcode fallback.
- **Intro/credit skip**: derived from Emby chapter markers (Emby has no Jellyfin MediaSegments API).

This includes the fixes worked through with @anthonycaccese in #153, each traced against a live Emby 4.9.5.0 server:

- **External/sidecar subtitles keep direct play**: the DeviceProfile now advertises `External` for text formats (Embed-only forced a burn-in transcode for subs that aren't muxed into the container).
- **Faster start**: mpv is handed only the selected subtitle sidecar. Preloading every embedded sub as a sidecar made Emby extract each on the fly (~27s for the first), stalling launch 30s or more even with subtitles off.
- **Dolby Vision to SDR transcode**: DV sources force a 1080p SDR transcode (mpv can't render the DV enhancement layer on macOS/Pi). See caveats below.
- Plus the earlier batch: episode-list 500 (`sortBy=AiredEpisodeOrder` is Jellyfin-only), direct play (Emby's default bitrate cap), Continue Watching (`mediaTypes` filter), and sign-out device de-auth.
- **List wraparound** across all Emby views, mirroring #174.

Relates to #153.

## AI involvement

Per CONTRIBUTING.md: the initial implementation was AI-generated (Claude Code; ported from the Jellyfin module), then built and tested locally by me. All playback fixes here were traced against my real Emby 4.9.5.0 server rather than the docs (the OpenAPI spec is incomplete), and the key ones were confirmed through the app UI.

## Testing (macOS ARM, Emby 4.9.5.0)

Verified in-app: password and Emby Connect sign-in, browsing, episode lists, Continue Watching, direct play, external-subtitle direct play, fast-start on titles with embedded subs, list wraparound, and Dolby Vision titles now playing (as a transcode).

Known limitations, called out honestly:
- **DV plays without subtitles**: this server 500s on any subtitle burn-in during transcode (SDR files too, a pre-existing transcode-path issue), so the sub is dropped on the DV transcode to guarantee playback. Follow-up: deliver subs as a sidecar over the transcode (also fixes quality-tier transcode plus subs).
- **Smooth 4K DV needs hardware transcoding on the server**: without it Emby software-transcodes at ~4 fps and stutters. Server-side, not the module.
- Not yet run on a Pi/CRT.

## Checklist

- [x] Sized with `root.sh` / `root.sw` (no hardcoded pixels), CRT overscan in mind
- [x] No tracking/analytics; only writes settings to the local data dir
- [x] Follows ARCHITECTURE.md patterns and CONTRIBUTING.md principles


anthonycaccese:
@freemchr awesome, I had more time today to work through this and just committed a few proposed updates

- [7f5053e](https://github.com/anthonycaccese/240-MP/pull/183/commits/7f5053ee9221334dd5d3c93a391f2dbb95a60695) - Updated logo.svg to monotone to it can take on accent colors correctly
- [a37430a](https://github.com/anthonycaccese/240-MP/pull/183/commits/a37430ad19e9739db3df15bfab3710a43cecbb1b) - Additional proposed fixes: added a logout SSL-host snapshot so device-deauth/token-revocation works on self-signed servers; implemented a single post-auth funnel in Root.qml (fixes double Libraries load + routes server_url through save_setting instead of direct config.json writes); added empty-list focus guards in ItemShow/ItemSeason.
- [37f26f8](https://github.com/anthonycaccese/240-MP/pull/183/commits/37f26f8c61ee20db3ee30299d5cc0a5d78a2e3a1) - Auth screen layout tweaks: Sign In / Emby Connect side-by-side with the "OR" separator and updated status text; Username/Password on one row; linear arrow/Tab navigation with wraparound (up/down, left/right and tab all linearly step through the form); grouped spacing between the input and action groups.

Auth view preview:
Screenshot 2026-07-23 at 8 52 36 PM

I tested things across MacOS and Raspberry Pi and all looks to be working well from my testing

I also set up a secondary test server to try out the emby connect flow and that worked end to end for me.

When you get time to review the proposed updates just let me know what you think and how it works in your testing.  If all pans out well I'd say we are a go for merge. 

Cheers

freemchr:
Thanks @anthonycaccese, went through all three commits and they hold up well. The SSL host snapshot in particular is a genuine catch: `logout()` clears `m_serverUrl` before the async `sslErrors` callback fires, so capturing the host by value at request time is exactly right. Same for routing `server_url` through `save_setting` instead of the backend writing `config.json` directly, which is what the contributing guide asks for anyway.

What I verified here on macOS against my server:

- Auth form layout, linear Tab order, and wraparound in both directions
- Monotone logo picking up the accent correctly
- Sign-out end to end (device deauth + token revoke), clean logs, no warnings
- The single post-auth funnel on **both** paths. Emby Connect is the interesting one: the server URL field isn't even filled in that flow, and `server_url` still persisted correctly from `get_server_url()`. That's your change doing real work.

**One thing I could not test, and would like your confirmation on.** My server runs plain HTTP (`:8096`, no TLS listener), so `QNetworkReply::sslErrors` never fires and the self-signed branch you fixed never executes on my setup. My clean sign-out proves the deauth/logout ordering, not the SSL fix. Did your testing hit a self-signed HTTPS server? If so I'm happy to call that one done.

**Emby Connect wants an email, not a username.** In practice the Connect flow only accepts the emby.media email address, even though the API parameter is `nameOrEmail` and our UI strings said "username or email". That mattered more than it sounds, because the old screen asked for credentials *before* you picked a method, so the shared field could only ever be labelled for one of the two flows, and the server URL box sat there for Connect even though `connect_authenticate` ignores it entirely.

I've pushed a follow-up that puts the method first:

- A "Sign-in Method" row (`Server (Local)` / `Emby Connect`) as the first focus stop, Left/Right to pick
- The credential field labelled Username or Email to match the chosen method
- Server URL row hidden under Connect rather than implying it's needed
- One Sign In button instead of two competing actions, since the method row already decided the flow

It keeps your layout language and your linear-nav model, it just reorders the decision. Both paths signed in end to end here. Very happy to drop or rework the whole thing if you'd rather keep the screen as you had it, you've just been through this file and I don't want to trample it.

I've also pushed fixes for the two follow-ups I'd originally flagged as separate work, since both turned out to be tractable:

**1. Libraries no longer double-loads on cold start.** `Root.qml` `Component.onCompleted` shows Libraries and then calls `check_auth()`, which emitted `authStateChanged()` on a successful re-validation even though the state was still "authed", so Root loaded Libraries again and rewrote `server_url` for nothing. `authStateChanged` is now only emitted on a real transition, seeded from the restored state in the constructor. Sign-out and revocation had the same doubling from the other side (Root navigated on `logoutComplete` / `authRevoked` *and* on the `authStateChanged` that follows), so routing is centralised on `authStateChanged`. Pre-existing, your `save_setting` line just made it visible in the logs.

**2. Subtitles now work over a transcode.** Selecting a subtitle on the transcode path meant no playback at all, because the profile advertised `Method=Encode` for every format and this server 500s re-encoding video with a subtitle painted on, for every subrip tested including SDR sources. Text subs now use `Method=External` on the transcode path, exactly as they already did for direct play. Verified against 4.9.5.0: with `External` the returned `TranscodingUrl` carries no `SubtitleMethod` and no `SubtitleStreamIndex`, so the video transcodes untouched and the subtitle is fetched from `/Videos/{id}/{ms}/Subtitles/{i}/Stream.srt`.

That alone still showed nothing, because the client was discarding the sidecar: `Player.qml`'s transcode branch hardcoded `--sid=no` and an empty sub-file list, which was correct back when the sub was burnt in. It now goes through `buildSubArgs` like direct play does. Image subs (PGS/VOBSUB) have no sidecar and are still burnt in, so they return `--sid=no` rather than an index into embedded tracks that an HLS stream doesn't carry. Confirmed on a forced transcode tier: playback starts and the subtitle renders and follows.

Worth a second pair of eyes on that last one in particular, since it changes the playback path rather than just the UI.

**One change I have deliberately not verified.** The Dolby Vision path forces an SDR transcode (mpv can't render the DV enhancement layer) and used to drop the selected subtitle on that re-request, purely as a workaround for the burn-in failure above. With text subs now riding a sidecar there is nothing left to work around, so it keeps the subtitle when the stream reports `IsTextSubtitleStream`, and still drops image subs since those are burnt in.

I could not exercise it: my test server is a Synology DS923+ with no iGPU, so it can only software-transcode DV at a few frames per second. The commit message says as much. If you have a DV title and hardware that can transcode it, that is the one thing in this PR that would benefit most from your testing rather than mine.

Cheers.

freemchr:
One more thing before you merge, @anthonycaccese.

Between us we've now added four commits on top of your three, and two of them contain code that neither of us has actually run. Everything else in here has been exercised end to end, so I'd rather name the exceptions than let them ride in on reasoning alone.

**1. Token revocation while browsing (`48522ac`).** `authStateChanged` is now only emitted on a real state transition, and Root's routing is centralised on it, with the `logoutComplete` / `authRevoked` handlers removed. I confirmed the cold-start case (Libraries no longer loads twice) and normal sign-out, but sign-out is invoked from app Settings, where the module's `Root.qml` isn't loaded, so it was `Component.onCompleted` that put me back on the Auth screen, not the path I changed. The path that actually depends on it is a **401 while you are inside the module**: expired or server-side revoked token, mid-browse. That should drop you on Auth.qml exactly once. If you can revoke the device from the Emby dashboard while the module is open, that is the case.

**2. Dolby Vision keeping a text subtitle (`65b40f6`).** Covered in the comment above: my server can't transcode DV at a usable frame rate, so I have no way to see whether the subtitle actually draws. A DV title with a text subtitle selected, on hardware that can transcode it, is the whole test.

Everything else I'd call verified: both sign-in methods, sign-out, the post-auth funnel on the Connect path, and subtitles over a forced transcode (confirmed rendering and following after a seek).

No objection at all if you'd rather merge and fix forward, it's your call and your repo. I just didn't want "ready to merge" to quietly cover two things I haven't stood behind.

anthonycaccese:
Great idea on the "tabbed" approach to the auth screen! Think that rounds things out nicely.  I did just push a small set of layout tweaks to it to give things a bit more space (after testing more on a CRT i prefer a bit more open spacing).  Let me know what you think...



The changes are pretty slight (essentially removing the "sign in options text", adding "or" and tweaking the group spacing a bit more.

> One thing I could not test, and would like your confirmation on. My server runs plain HTTP (:8096, no TLS listener), so QNetworkReply::sslErrors never fires and the self-signed branch you fixed never executes on my setup. My clean sign-out proves the deauth/logout ordering, not the SSL fix. Did your testing hit a self-signed HTTPS server? If so I'm happy to call that one done.

Yes one of the tests I set up was self-signed and the flow worked well in testing so thinking this is ok to call this done (and we'll see if we get any additional feedback after)

> Subtitles now work over a transcode

Tested and looking good on my end too.

> Dolby Vision keeping a text subtitle

I don't currently have Dolby Vision assets to test with myself but am ok to call this something we can work through later if it comes up as an issue

> Token revocation while browsing

I verified this by having the module open and revoking 240-MP via the devices tab on my test Emby Server.  The next selection event in app was caught correctly, threw an error and backing out got me to the Auth Screen.   So for me this is working as expected, especially for a niche hobby project =)

Related to this we should aim to write up a POV for auth to have on the wiki.  

I wrote this for Plex: https://github.com/anthonycaccese/240-MP/wiki/Module:-Plex#authentication-model 

And I think for Emby; call outs for things like "how are passwords handled" will be important to document.  Reading through the auth approach and having the source for it be open so it can be audited is a good start and then also being honest about this is a hobby meant for local trusted networks I think will help people make a decision on how they want to use it or not .  the points I wrote under the "What does that mean for you?" header I think are 1:1 but let me know what you are thinking for a write up like this for Emby.

freemchr:
Thanks @anthonycaccese, that closes out everything I had open.

- **SSL self-signed**: you hit that path on a real self-signed server, so I'm happy to call it done. That was the one thing my setup physically couldn't exercise (plain HTTP, no TLS listener), so your confirmation is the only thing that could have closed it.
- **Token revocation while browsing**: revoking via the devices tab with the module open is exactly the case I couldn't reach from app Settings. Good to have it confirmed.
- **Subtitles over a transcode**: matches what I saw here.
- **Dolby Vision + subtitle**: agreed, park it. It needs a server that can transcode DV at a watchable frame rate, and neither of us has one. If it surfaces, it'll surface with someone who can actually test it.

On the layout tweaks in `343b4c2`, the extra breathing room reads better, and dropping the "sign-in options" line for the "OR" between the two buttons is cleaner than what I had. To be straight about it: I've read the diff rather than run it, so that's an opinion on the code and on your CRT testing, not a second test. I'll say something if anything looks off next time I build. One cosmetic nit only: the `MethodOption` width comment says `//218` but `root.sw * 0.321875` works out to 206. Comment, not behaviour.

**So: go for merge from my side.**

---

### Emby wiki write-up

I did the auth POV you asked for, and ended up writing the whole module page rather than just that section. It needs the page around it anyway, and I followed the Jellyfin page's shape with the Authentication Model section in the voice of your Plex one.

It's drafted on my fork's wiki so it's readable in context and doesn't put anything on yours before you've seen it:

**https://github.com/freemchr/240-MP/wiki/Module:-Emby**

Copy it across, edit it, rewrite it. It's yours. Everything in it is read out of `EmbyBackend.cpp` rather than written from memory.

Alongside auth it covers: both sign-in flows, direct play vs the transcode tiers (including the default-bitrate-cap trap, since that one looks like a codec problem and isn't), the subtitle behaviour on both paths, the fast-start fix, chapter-marker intro/credit skip, the shelves, and a Known Limitations section.

You said the "What does that mean for you?" points are 1:1, and mostly they are. But **three things genuinely don't carry over from Plex**, so I'd rather flag them than have them look like careless copying:

**1. The Plex module never sees your password. The Emby module does.** Plex is PIN auth via plex.tv/link. Emby's `/Users/AuthenticateByName` takes the password directly in the request body. The page says plainly that 240-MP never writes it to disk and doesn't keep it past the request, and then makes the call-out that follows from that: if the server is plain HTTP, which is the common LAN default and is what mine runs, the password crosses the network in the clear. That's the Emby API rather than anything we add, and it's identical in the official apps and the web UI. But it felt wrong to write a page about how we handle your password and leave it out. It's framed with the two ways out: HTTPS, or Emby Connect, where it only ever goes to `connect.emby.media` over verified TLS. Very happy to soften the wording if it reads alarmist to you. I'd just rather not drop the fact.

**2. Emby access tokens don't expire.** Your Plex text leans on the JWT having a bounded life and a rolling refresh. Emby's token stays valid until something revokes it, so I flipped that paragraph instead of copying it: no bounded life makes the `Dashboard > Devices` kill switch matter more, not less. Which ties straight into what you just verified.

**3. Your self-signed fix is documented.** The page names the exact four errors we ignore, that it only applies to the configured server host, and that `connect.emby.media` is never relaxed. Plus the honest trade-off: encrypted but not authenticated, so it's not proof against someone already on your network. Same bargain every Emby client makes to support self-signed setups, but worth writing down.

Also documented: Connect takes the emby.media email only (never a username), Connect drops out of the path entirely after the token exchange, and the sign-out ordering including why the device id is regenerated each time.

PR #184, opened 2026-07-23
feat: Live TV (IPTV) module

freemchr:
## Summary

Adds a self-contained **Live TV (IPTV)** module (`modules/iptv/` plus `src/modules/iptv/`, one `registerModule` call). Channel-surfing for the VCR concept: point it at an M3U playlist and browse/watch. Ships disabled by default.

- **Playlist**: an M3U from a remote URL or a local file. Channels are grouped by `group-title`, with an "ALL CHANNELS" shelf first and a "CHANGE PLAYLIST" entry so a bad URL can be fixed without leaving the module.
- **Guide (optional)**: an XMLTV URL gives a now/next line and a progress bar per channel, matched to channels by `tvg-id`.
- **Playback**: streams are handed straight to mpv untouched (HLS/TS/UDP). No auth, transcode, or DRM, in keeping with browse-and-hand-off.
- **List wraparound**: the channel and group lists follow the app-wide pattern from #174 (added here since these views are new).

Relates to #153.

## AI involvement

Per CONTRIBUTING.md: the initial implementation was AI-generated (Claude Code), then built and tested locally by me against a real playlist and guide. The wraparound was added and the module re-verified in this pass.

## Testing (macOS ARM)

Verified against a live Melbourne free-to-air feed from Matt Huisman's free over-the-air service (a good, legal test source if you want to try it yourself):

- Playlist: `https://i.mjh.nz/au/Melbourne/raw-tv.m3u8` (167 channels)
- Guide: `https://i.mjh.nz/au/Melbourne/epg.xml`

Checked M3U parse and grouping, the channel list, playback (streams hand to mpv), and the EPG (126 channels matched, ~15.8k programmes) with the now/next lines confirmed against actual broadcast times. The guide URL serves a 302 redirect, which Qt's network stack follows, so the guide still loads. `i.mjh.nz` has equivalents for other AU/NZ regions and a few other countries if you want a different region. Not yet run on a Pi/CRT.

## Checklist

- [x] Sized with `root.sh` / `root.sw` (no hardcoded pixels), CRT overscan in mind
- [x] No tracking/analytics; only writes settings (playlist/guide URLs) to the local data dir
- [x] Follows ARCHITECTURE.md patterns and CONTRIBUTING.md principles

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ add_executable(240mp
     src/modules/local_files/LocalFilesBackend.cpp
     src/modules/plex/PlexBackend.cpp
     src/modules/jellyfin/JellyfinBackend.cpp
+    src/modules/iptv/IptvBackend.cpp
     src/modules/ambient_mode/AmbientModeBackend.cpp
     src/modules/youtube/YouTubeBackend.cpp
     src/modules/nfc_reader/NfcReaderBackend.cpp
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Watch on YouTube: https://youtu.be/r-gylGDoELY
 ### Live TV Module (IPTV)
 - Designed for CRT navigation (simple, fast, list browsing)
 - Point it at an M3U playlist (remote URL or local file)
+- Optional XMLTV guide URL for now/next info — current programme, a progress bar per channel, and a now/next info bar on tune-in
 - Channels grouped by category (`group-title`), plus an "All Channels" shelf
 - Streams handed straight to mpv (HLS/TS/UDP) — no auth, transcode, or DRM
 - Change the playlist any time from the "Change Playlist" entry
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 240-MP is a retro VCR style frontend to play content on [Raspberry Pi](https://github.com/anthonycaccese/240-MP/wiki/Hardware-Testing) (preferably hooked up to a CRT TV).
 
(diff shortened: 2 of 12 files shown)


freemchr:
@anthonycaccese no rush on this one, but flagging it in case it slipped past: #184 (Live TV / IPTV) hasn't had eyes on it yet.

It's independent of the Emby work, so it can be looked at whenever suits. It's a self-contained module (`modules/iptv/` plus `src/modules/iptv/`, one `registerModule` call), ships disabled by default, and touches no existing module code, so the blast radius is small. Playback is pure browse-and-hand-off to mpv: no auth, no transcode, nothing to negotiate.

If you want a quick way to try it, the free-to-air feed I tested against is a good legal source:

- Playlist: `https://i.mjh.nz/au/Melbourne/raw-tv.m3u8` (167 channels)
- Guide: `https://i.mjh.nz/au/Melbourne/epg.xml`

`i.mjh.nz` has the same pair for other AU/NZ regions and a few other countries if you'd rather test something local to you.

The one real gap is that I've only run it on macOS. It's never been on a Pi or in front of a CRT, and given it's a grid of channel rows with a now/next progress bar, overscan and readability at that size are exactly the things I can't check from here. That's the part where your setup would tell us more than mine can.

Happy to make changes to it the same way we did on #183.

anthonycaccese:
Thank you, I didn't forget about this =) just have been focused on other items.  

I have started using this on my CRT and am going to propose some view updates to better account for overscan over the weekend.   But good news is it does function end to end, there is nothing I am seeing the approach you took that makes it not compatible with CRTs at a browse to playback level.

I had thought you also had a CRT so sorry for my mistake on that, with that in mind i'll aim to do most of my work/testing on this on the CRT.

freemchr:
No apology needed @anthonycaccese, and that's genuinely good news, that it browses to playback end to end on a CRT without anything structural getting in the way. That was the part I couldn't check from here, so it's the part I most wanted to hear.

Overscan and readability are all yours. You've got the display, and you know what actually reads at that size far better than I can reason about it from a macOS window, so propose whatever you think it needs and I'll follow your lead rather than second-guess it. Same working pattern as #183, which went well.

The two things most likely to need it, if it helps aim: the channel rows are dense, and the now/next progress bar is the smallest thing on screen.

No rush at all. #183 is the one with momentum. Once that lands I'll flip this out of draft, since the only reason it's draft is the one-open-PR cap for non-collaborators.

anthonycaccese:
Just a note to say i've been heads down working on a weather module so have not spent the time i originally planned to with this yet.  I plan to wrap my changes for weather by the weekend and will look to shift focus to this (and other open PRs) then.

PR #188, opened 2026-07-25
feat(jellyfin): change subtitle track during transcode

ToTheXtreme64:
## Summary

Allow changing subtitle track during playback in the Jellyfin module during transcoding. 
## AI involvement

Used to check my work

## Testing

Tested on MacOS and works perfect from what I have tested. 

## Checklist

Probably needs some further testing to make sure I didn't break anything. Also having a menu to switch subtitle and audio tracks instead of just cycling through them might be a good idea.

### DIFF
--- a/modules/jellyfin/views/Item.qml
+++ b/modules/jellyfin/views/Item.qml
@@ -313,11 +313,7 @@ FocusScope {
                 }
             }
         }
-        if (jellyfinBackend && typeof jellyfinBackend.set_last_track_langs === "function") {
-            jellyfinBackend.set_last_track_langs(aLang, sLang, aLangIdx, sLangIdx)
-        } else {
-            console.log("[Item] Skipping set_last_track_langs: backend method not available")
-        }
+        jellyfinBackend.set_last_track_langs(aLang, sLang, aLangIdx, sLangIdx)
     }
 
     function applyLanguagePreferences(d) {
--- a/modules/jellyfin/views/Item.qml
+++ b/modules/jellyfin/views/Item.qml
@@ -313,7 +313,11 @@ FocusScope {
                 }
             }
         }
-        jellyfinBackend.set_last_track_langs(aLang, sLang, aLangIdx, sLangIdx)
+        if (jellyfinBackend && typeof jellyfinBackend.set_last_track_langs === "function") {
+            jellyfinBackend.set_last_track_langs(aLang, sLang, aLangIdx, sLangIdx)
+        } else {
+            console.log("[Item] Skipping set_last_track_langs: backend method not available")
+        }
     }
 
     function applyLanguagePreferences(d) {
--- a/modules/jellyfin/views/Player.qml
+++ b/modules/jellyfin/views/Player.qml
@@ -382,12 +382,22 @@ FocusScope {
     function doStartPlayback(offsetMs) {
         var jfToken = jellyfinBackend.get_access_token()
         if (isTranscoding) {
-            // HLS manifest bakes in the selected audio, and the chosen subtitle is
-            // burned into the video — so there's no soft sub track for mpv to pick
-            // (subTrack -2 = --sid=no, a no-op when nothing soft exists).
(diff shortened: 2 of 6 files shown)


anthonycaccese:
@ToTheXtreme64 overall this is a good find and nice QOL addition for Jellyfin running with transcoding instead of direct play.

But when testing through this morning i found the OSC changes caused major regressions across all other modules.  For example: the original change impacted playlist support so navigating through items in a playlist would not work in local files or youtube and for Plex transcodes it added a SUBTITLE button that did nothing when pressed (direct play was unaffected, since the .m3u8 check didn't match). It also broke screensaver launching during Jellyfin transcode playback, since the new --script-opts= wiped the list MpvController had already built (screensaver_timeout, and hide-crop on Pi 3)

So thats the reason for the commit i just pushed which refactors the approach to still accomplish what you were after while keeping the radius of the fix to Jellyfin and not other modules.

Some lessons this helped me work out to keep in mind for future work...
- scripts/mpv-osc.lua is shared (Every video module loads it) So a change there needs to be justified against Plex, Local Files, YouTube and NFC, not just the module being worked on. 
- aim to prefer sending an explicit signal from the app vs inferring context inside the Lua. The Lua should try as much as possible to act on signals from the app vs inventing its own internal logic.  For example... Sniffing .m3u8 in the mpv path reads like a safe Jellyfin check but it matches any HLS source. The `hide-crop` opt (for pi3) is a good pattern to copy: the app knows the situation and states it, the script doesn't guess.
- don't overload an existing script-message. Reusing skip-segment avoided touching MpvController (which I understand why you avoided that so its all good), but a second meaning on one channel needs a disambiguator on the QML side, and that ambiguity became a collision with intro-skip.
- extraArgs are appended after the app's own mpv flags. Any list-valued option (--script-opts especially) replaces rather than merges so please use the -append form. Worth a one-line check of what MpvController already passes before adding a flag.
- mpvController.stop() is asynchronous and runs the normal end-of-playback path. Anything that quits mpv intending to relaunch has to sequence through onPlaybackEnded, or it races the exit; the existing direct-play-failure retry in the same file is the working template.

One bonus this helped us work through though...
The app already fetched text subs as sidecars and handed them to mpv; but our device profile only advertised Embed, so the server refused direct play for any external track and fell back to a burn-in transcode. With the changes in this commit we now advertise External too for text formats (srt/ass/ssa/vtt/mov_text). External image subs (PGS/VOBSUB) still transcode as there's no sidecar for those.

Please run a test end to end on your end for the modules you have access to, i'll do another scrub on my raspberry Pi (as I've only tested on MacOS so far).  

The tests i've run through (if it helps to see):
playlist </> in Local Files or YouTube
Plex transcode subtitle toggle
Jellyfin direct play with an external .srt
Jellyfin transcode subtitle switch

Thank you and let me know

ToTheXtreme64:
@anthonycaccese Thank you so much for the insight, I'm new to working on a project like this and am learning as I go. The changes you made look great to me, but I have just noticed that while direct playing with a SUBRIP or DVDSUB embedded in the file a transcode is forced from the server with the message "The video's bitrate exceeds the limit" if the subtitle is selected. External subtitles work fine. Thank you again!

anthonycaccese:
no worries and all good, we're all learning so I am more glad its helping you do that than anything else (jellyfin is cool and all too 😊).  for "The video's bitrate exceeds the limit" wondering if maybe Jellyfin has a cap or setting on what max bitrate it allows direct play on before sending a transcode... if so maybe something we can set?  I don't daily drive Jellyfin so there are a lot of aspects of it that I am not fully versed on yet.

ToTheXtreme64:
I don't think it's a problem with the bitrate. I am able to direct play a 80GB 4k remux fine with the Jellium Desktop client on my PC. I am able to switch between all the subtitles without transcoding. This is definitely a separate issue as I was able to confirm the issue occurs before this pull request. It has something to do with the embedded subtitles. The video will direct play with no subtitle selected then switch to transcoding after selecting a subtitle track.

PR #192, opened 2026-07-25
fix(jellyfin): direct play high bitrate files

ToTheXtreme64:
## Summary

Allows for videos with high bitrates to be direct played in the Jellyfin module. Using a cap of 0 causes inconsistent direct play behavior, setting a very high cap seems to fix the issue. 
## AI involvement

Not used
## Testing

Tested on MacOS with a few very high bitrate movies and everything direct played fine with seamless subtitle switching. 

## Checklist


anthonycaccese:
1000 Mbps!

giphy

Nice find =)

PR #211, opened 2026-08-01
feat(jellyfin): change audio track during transcode

ToTheXtreme64:
## Summary

Allows the user to change the audio track while transcoding in the Jellyfin module. Uses a very similar implementation as my earlier PR for subtitle switching. 
## AI involvement

Used to check my work
## Testing

Tested on MacOS with my Jellyfin server

## Checklist

### DIFF
--- a/modules/jellyfin/views/Player.qml
+++ b/modules/jellyfin/views/Player.qml
@@ -33,7 +33,8 @@ FocusScope {
     property bool   pendingRetryTranscode: false
     // Set when the user changed the subtitle mid-transcode: mpv is quitting so we
     // can re-request the stream with a different burned-in track (not a real stop).
-    property bool   pendingSubtitleSwitch: false
+    property bool pendingSubtitleSwitch: false
+    property bool pendingAudioSwitch: false
     property string carryAudioLang:     ""
     property string carrySubLang:       "__off__"
     // Full stream metadata used to disambiguate when several streams share a language.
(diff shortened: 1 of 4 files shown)


anthonycaccese:
This looks great @ToTheXtreme64, great all on mirroring your subtitle work too.

adding my commit message here to make it easy to discuss if you have questions on the updates I am suggesting...

It's likely overkill but I chose to refactor  pendingAudioSwitch/pendingSubtitleSwitch into a single pendingTrackSwitch flag. Its def an edge case; with 2 seperate flags there was a small chance of selecting to switch audio and then trying to switch subtitle before that finished that would cause the player to get into a weird state. It was very difficult to trigger but thought this could be a good one to add as one flag makes will make that state unreachable for good.

With transcoding turned on getting to the audio "off" state felt odd (because it tore down the stream to switch to an empty track which you would then need to do again to get to the actual next audio track) so I elected to drop the "off" state from the audio cycle when transcoding. Direct play is untouched and keeps mpv's own no-audio step from `cycle audio`.

Added a gate for audio-cycle on audioStreams.length similar to how sub-cycle works; this will help in the off chance where a video actually has no audio tracks to display.

I tested these changes on my own local instances (MacOS and RPi 4) and its working well.  but let me know what you think.

ToTheXtreme64:
Everything sounds good to me. I kept the off state to mirror what was already there, probably better to just get rid of it. Maybe also get rid of it when direct playing also.

PR #212, opened 2026-08-03
feat: TV Mode module

Incluwin:
Refer to the readme for features: https://github.com/Incluwin/TV-Mode-240-MP/blob/crt-fork/README.md

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,12 @@ add_executable(240mp
     src/modules/youtube/YouTubeBackend.cpp
     src/modules/nfc_reader/NfcReaderBackend.cpp
     src/modules/weather/WeatherBackend.cpp
+    src/modules/tv_mode/TvBackend.cpp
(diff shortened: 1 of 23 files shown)


anthonycaccese:
Hi @Incluwin, I am sorry for my delay on this.  I've been playing with this in a local branch and I think you've done great work.  

Though I think because of how focused it is on creating a TV channel experience (and how you've automated that setup through folders) that it would make sense as a standalone application instead of as a module in 240-MP (kind of thinking of 240-MP as the vcr/dvd player you attach to your TV and your work here as the other side of that).   And with the new scripts module (https://github.com/anthonycaccese/240-MP/pull/227) a simple launch script could be created to open directly into a channel playing experience I think.

I'd be glad to help you pursue a standalone app if thats something you'd be interested in but if not then I think the fork you have is great to stand alone as well so please feel free to promote it.   Thank you for bringing the idea here.

PR #227, opened 2026-08-09
feat: scripts module

### LINKED ISSUE #128: [Idea]: run .sh scripts from menu
### Area

App shell / general

* A menu option/module that allows users to run their own .sh scripts from a local folder directly from 240-MP
* A Favorite option to save favorite scripts to the main module menu to show alongside other modules like applications, maybe even toggle autoboot like other modules
* Other options such as terminal display toggle could be included

### Why is this useful?

This would allow users to run their own .sh scripts to run programs like FieldStation42,  update things like yt-dlp, or really anything, all directly from 240-MP's remote/controller friendly interface. There will probably be issues with entering passwords for some scripts, but I imagine that could be addressed by including a warning that a keyboard may be required for some scripts. I know it might not be the best solution and kinda goes against the hands-off approach, but I think the feature is worth it and shouldn't be a problem for most people.

A lot of this idea stems from me wanting to boot FieldStation42 on my Pi without SSH, and still have access to my library of videos, YouTube, and hopefully Retroarch later down the line. I've thought about whethet it could be ised to autoboot FS42 when I boot up the Pi, but that's possible other ways so it's not a priority. Either way, this is a feature I think would be really useful, and I hope it's considered!

anthonycaccese:
## Summary

Closes: https://github.com/anthonycaccese/240-MP/issues/128

Adds a scripts module that allows you to point a folder of `.sh` files and run any of them from the menu with a remote.  Favorites set in the module are displayed on the main menu alongside other modules.  Basically enables 240-MP to
act as a front end for other apps on the device that aren't direct modules (think things like launching FS42, Retroarch, an update script for YT-DLP, etc...)

Supports the following modes:

| Mode | Screen during run | Returns |
|---|---|---|
| `console` | 240-MP keeps the display and outputs the stream of the script into Console.qml | holds `EXIT n` so you can read the output of the script.  and then just press the back button to get back to the list of scripts |
| `takeover`, launched script draws something to the display | The launched script paints over a hand-off card on Takeover.qml | yes, on clean exit |
| `takeover`, launched script draws draws nothing to the display | The hand-off card on Takerover.qml stays displayed until display is restored | yes, on clean exit |

Per-script metadata lives in sidecar `.txt` files (`foo.sh` → `foo.txt`) to allow you to name, favorite and set options for how the script will be run.  And the sidecar files are auto-generated when a new script is found in the folder so you can add metadata without needing to remember the format.

This work also drove a refactor of the display handoff logic that I had in MpvController as the Script module needed similar logic to handle display handoff for scripts.  So instead of duplicating that logic its not centralized in `DisplayHandoff ` which both MpvController and ScriptsLauncher have been updated to call. This also addressed a long standing bug with pkill / systemctl stop that orphaned mpv, leaving music or fullscreen video running with no app.  main.cpp now handles SIGTERM/SIGINT/SIGHUP and routes to quit() so destructors actually will actually run now for those cases.  

## AI involvement

Used Claude Opus 5 to work through the MpvController > DisplayHandoff refactor, plan thh architecture for the scripts module to handle console and takeover modes, update AppCore to enable a generalized way to enable modules to add items to ModuleList (scripts is the example for this with favorites), debug display handoff for the RPi in headless mode and work through the sidecar metadata approach.  I've reviewed the code, documentation and manually tested through each commit on RPi and MacOS

## Testing

Pi and MacOS verified on hardware across the build, that said as this PR comes with a large refactor of DisplayHandoff I'll be doing an alpha release and full test cycle (both fresh install and update on RPi, MacOS and SteamOS before setting this for ready to review) 

## Checklist


anthonycaccese:
Regression testing is going well and i've not found any playback conflicts/regressions on RPi4, MacOS, SteamDeck in each of the modules.  I've tested both autostart and dev set ups as well.  Tested against the current alpha

That said I did find one thing in the Takeover logic that I didn't work as I had planned.  I added a way to exit out of a script by pressing [ESC] and for takeover scripts that key is still being captured in the takeover application so doing something like pressing back while in RetroArch tells 240-MP to stop Retroarch (not that useful when the back button is also the "A" button on a controller 😄).  

So I am thinking through options and will push a commit to this PR once I have something I think could work - then will do another round of regression testing.  I'll also reach out to folks with the next alpha to get a few more eyes on this to check if anything else is missing/not-working ahead of a merge.  To help with that I'll write up a wiki page for this as well as instructions for updating to an alpha release.

Progress for sure, just stepping through it bit by bit =)

anthonycaccese:
Fixes for takeover input are working well.  I added details on how to test the alpha here: https://github.com/anthonycaccese/240-MP/issues/128#issuecomment-5280408706

anthonycaccese:
testing looks good from the latest commit/build on steamOS so going to merge this as that was my last bit of regression testing =)

I don't have a fieldstation 42 instance to test with but I have tested takeover script running on all device targets using RetroArch so I think it should hold up fine.  If not we can always fix it in a future release.  

Merging now.

PR #243, opened 2026-08-15
feat(mpv): Allow holding key to skip repeatedly

ToTheXtreme64:
## Summary

Can now hold left or right while on the playback bar on the osc to repeatedly skip. 
## AI involvement

Not used
## Testing

Tested on MacOS

## Checklist


ToTheXtreme64:
I just noticed completely separate from this change that sometimes when playing transcoded media, if you try to skip to the beginning of the video it can get stuck and never be able to reach the beginning. It takes a few tries but it is there, not sure how to fix it but thought it was worth pointing out.

anthonycaccese:
awesome, great addition @ToTheXtreme64

PR #250, opened 2026-08-21
feat(plex,nfc reader): play plex content via nfc cards

anthonycaccese:
## Summary

This PR adds a few core things:
1) A generalized foundation to enable the NFC Reader module to hand off playback of mapped cards to other modules (built with Plex as the first test)
2) A pattern for writing NFC cards from within a given module.  The pattern for this is to only show the option if 2 things are true... the NFC module is enabled and an NFC reader is plugged in.  This PRs provides an integration for Plex that we could look to model for Jellyfin, Emby, etc...

### What a card holds with this change

Existing URL and path based mappings are unchanged.  This PR adds mapping for Plex guid and an optional 3 line to define playback mode

A Plex card stores the item's `guid` rather than a ratingKey, so it survives library re-scans and keeps resolving if the item moves between servers — which a physical card on a shelf needs. Cards can point at a movie, episode, season or whole show.

Show and season cards take an optional mode on a new third line of the tag file: sequential (next unwatched, so tapping again picks up where you left off) or shuffle. Shuffle plays jukebox style and reports no progress back to Plex, so leaving one running doesn't mark episodes watched or fill Continue Watching. The mode line is deliberately generic rather than Plex-specific, so `.m3u` and YouTube playlist cards could use the same slot later if we want.

### Not in scope (and a pattern i'd like to keep in mind if we NFC card mapping to other modules...)

Cards resolve against whatever server and user is currently selected, they do not carry server/user changes.  A miss is a plain "not found on this server" error.  Switching servers would add complexity in persisting config as a side effect of a card tap, and auto-switching a Home profile would be a PIN bypass in physical form so i'd like to keep the pattern as... "Cards inherit the current settings of the module they are passing do; they do not change module settings"

## Video of how this works

https://github.com/user-attachments/assets/4682d47b-6b14-4f38-98bd-4c21bf19d25c

## AI involvement

Used to structure the handoff logic from NFC Reader to Plex and model the pattern that we can reuse for handoff to other modules.  I reviewed the logic flow and wrote the QML changes to integrate writing cards into the Plex module as well as the NFC writer component.

## Testing

Verified on macOS and RPi across movie, episode, season and show cards, both modes, and the failure paths (wrong server, signed out, module disabled, unresolvable reference). Card writing and playback both confirmed with real card taps.

## Checklist


anthonycaccese:
@jmcerrejon, thought you would be interested in this as an idea for enabling NFC reader module to hand off playback to other modules in 240-MP.  

I was creating cards for my favorite movies and thought it would be helpful to just point them to my existing plex library (which would mean I could get things like continue watching, play history for free) so this PR is the result of that =)

Its mostly generalized so we can also explore similar hand off paths for jellyfin, emby etc.  I am not sure if you have a Plex server to try it out but no worries if not as I mainly wanted to share it with you as an idea on possibilities.

jmcerrejon:
I don't have Plex right now but 
I'll check it the PR, sure! Great addition btw 🥳

anthonycaccese:
alpha for testing: https://github.com/anthonycaccese/240-MP/releases/tag/v2026.08.23-PR250-alpha

src/modules/nfc_reader/NfcReaderBackend.cpp
```diff
@@ -398,6 +449,82 @@ void NfcReaderBackend::setModuleActive(bool active) {
...
+        if (prevPath != path && QFile::exists(prevPath)) {
+            if (QFile::remove(prevPath))
+                qDebug("[NfcReader] Removed previous tag file: %s", qPrintable(prevPath));
+            else
+                qWarning("[NfcReader] Could not remove previous tag file: %s", qPrintable(prevPath));
+        }
+    }
+
+    QFile file(path);
+    if (!file.open(QIODevice::WriteOnly | QIODevice::Truncate)) {
```
jmcerrejon, line 524: So, if you need two or more cards with different UIDs (for example, Dune 1984, Dune 2021, or Dune TVShow (2K), what file name has each card? Is it possible to have 3 Dune.txt files?

modules/plex/views/Player.qml
jmcerrejon: maybe if trackProgress and partId ?

trackProgress: false only suppresses timeline updates. When shuffle autoplay advances to the next episode, advanceToEpisode() still calls set_audio_stream() and set_subtitle_stream(), which persist the selected tracks in Plex for every episode played.

I think this conflicts with the card-play contrat that NFC playback should not mutate Plex state.

jmcerrejon (commented):
Not completely sure about that. Check it out and take it into account or not. 😅

src/modules/nfc_reader/NfcReaderBackend.cpp
anthonycaccese: Great catch, thank you!  Added a check here now to append `(#)` to the filename if a name that is about to written matches with the name of another tag (specifically in the case with a different card ID).  Also made some QOL updates to the Plex end of things to append Year to the file names that are generated.

modules/plex/views/Player.qml
anthonycaccese: Yup I missed that, fixed now to add a guard on `trackProgress && partId`

anthonycaccese:
Thank you, I agree on both points and added [62335d6](https://github.com/anthonycaccese/240-MP/pull/250/commits/62335d64ce674aabf9556f03d0dc8ac9212f99d4) to address your feedback

Tested across all targets again and name conflicts are working well now as well as the guard against persisting audio/sub choices back to plex on shuffle.  Merged!

Updated alpha: https://github.com/anthonycaccese/240-MP/releases/tag/v2026.08.23-PR250-alpha

PR #252, opened 2026-08-22
fix(weather): weather unit request race

snkinard:
I initially observed this behavior around the time when the (AWESOME) weather module was first released. I finally got around to attempting to track down what was happening. It's easily reproducible with mocked network calls but intermittent when fully connected to the real third party dependencies.

## Problem
Changing Weather units in settings starts replacement requests, but if there's a request that was sent earlier its possible for the earlier request to finish afterward and overwrite the newer results. This can leave Celsius data visible after selecting US units, show an `°F` Other Locations heading over Celsius values, or mix stale forecast units with observations formatted using the current setting.

## Fix
- track primary Weather and Other Locations requests independently
- when a new request starts, supersede older requests in that same family
- ignore callbacks from superseded requests so the latest unit selection always wins

## Testing
- a deterministic fake-network harness reproduced the race and verified the fix for both request families; it was removed afterward because it was over 300 lines of code and the repository has no automated test infrastructure
- human-verified on macOS against live Open-Meteo and NWS: switching US → Metric → US kept Current Conditions, Extended Forecast, and Other Locations values and labels consistent
- verified on a Raspberry Pi 3 Model B Plus: after selecting US units, late Metric replies did not overwrite primary Weather or Other Locations data

## Scope
- `stop()` request invalidation is not changed
- cached NWS observation mixing does not have a separate regression case yet
- settings persistence was not retested because this change does not modify configuration storage or the `units` setting; it only rejects stale network replies
- live services were used only for the isolated macOS smoke test

## AI involvement
Codex generated the implementation and temporary reproduction harness under the human author's direction. The normal application build and behavioral verification passed on macOS and Raspberry Pi, and Sam's macOS visual smoke test passed. The final patch has received a human code review from the author.


anthonycaccese:
thank you! i should get time to review this tomorrow

snkinard:
All good and no rush!

anthonycaccese:
Great catch with this @snkinard, changes look good and working well on MacOS/RPi in testing. Merging.

snkinard:
Thank you, @anthonycaccese! It happened the very first time I ran weather module and I thought at first I had maybe inverted my lat long coordinates or something. I maybe only saw it once or twice since then which is why I dragged my feet getting around to investigating it.

snkinard:
@anthonycaccese pardon my question if it's dumb (I am a middle manager in my day job, after all): is the PR/ branch/whatever title pattern you use part of some standard `fix(weather): <PR TITLE>`? I'm curious and probably want to start following that standard myself if so

snkinard:
also do you have a flipper zero?

anthonycaccese:
yup the convention I am aiming for is more along the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) style: 
`<type>(<scope>): <subject>`

`<scope>` I think aligns well to module (like weather) or app concept (like mpv) but still working through cases there as they come up

I don't have a flipper zero, have been interested in checking it out but just haven't opened that door yet.

snkinard:
> I don't have a flipper zero, have been interested in checking it out but just haven't opened that door yet.

image

Been working on a small thing on the side. Was waiting to bake it a little more to put it in front of you but I just can't help myself

anthonycaccese:
super cool!  would def be interested in seeing more about how that works when you're ready

PR #262, opened 2026-08-25
Add AirPlay module: AirPlay 2 audio receiver

ly-man-do:
Adds a new **AirPlay** module (audio only — no video/screen mirroring) that lets an iPhone/iPad/Mac stream audio directly to a 240-MP device via AirPlay 2, with a themed Now Playing screen (album art or a generic cassette icon, title/artist/album, sender device name).

## What's included
- `AirPlayBackend` launches/owns `shairport-sync` as a subprocess bound to the module screen's lifetime (open = start, close = stop) and republishes its metadata-pipe output as QML properties.
- Auto-restarts on a mid-session crash (capped at 3 attempts, with a persistent on-screen error if it keeps failing) rather than silently sitting on the idle screen forever.
- Audio Output setting only offers `plughw:` ALSA devices — raw `hw:`/`dmix:`/`default` devices were found to crash `shairport-sync` outright on some hardware (confirmed on Raspberry Pi HDMI output).
- `scripts/setup-airplay.sh` installs `nqptp` (systemd service) and `shairport-sync` (`--with-airplay-2 --with-metadata`), preferring a prebuilt arm64 binary from the latest release and falling back to a source build; includes an audio self-test across all output devices.
- `.github/workflows/release.yml` gains a `build-airplay-deps-linux-arm64` job (native arm64 runner) publishing the prebuilt binaries used above.
- **Raspberry Pi only**: AirPlay 2 depends on `nqptp`, which only supports Linux/FreeBSD — no macOS build path.

## Testing
Tested on real hardware: a Raspberry Pi 4 (HDMI) and a Raspberry Pi 3B (composite output), both with a live AirPlay 2 connection from an iPhone including metadata and artwork. Also includes dev/test tooling (a fake `shairport-sync` fixture + smoke test, no real hardware required) and full wiki documentation.


anthonycaccese:
I appreciate you building this @ly-man-do but as this is Raspberry Pi only its not something i'd bring in, I am focused currently on modules that work cross device.  If that changes in the future i'll post details.  Thank you.

ly-man-do:
@anthonycaccese I see, thanks for the info. I can explore adding macOS support. I also noticed that Local Files does not have an audio player so parts of this could be reused there. Would it be better to bring these up in the discussions section first?

anthonycaccese:
> Would it be better to bring these up in the discussions section first?

Yes that works well, open discussion in the idea section: https://github.com/anthonycaccese/240-MP/discussions/categories/ideas and we can talk through approach ideas there for sure.  

Local Files music support is def something on my mind to look into so that could be a great first one.  Thank you

PR #276, opened 2026-08-30
Feat/virtual channels

munderwoods:
Hey I had a week off from work and decided to work on this thing I'm passionate about. Thanks a bunch for the great project! I wanted to add this functionality to suit my specific needs, but it would be pretty lame to just use it for myself and not send it back to you. Obviously feel free not to use it, I really just made it for myself. It's 100% AI generated, but my standards are pretty stringent since I'm a software developer. That being said, there are going to be some bugs since it's pretty complex, and I don't have an emby or jellyfin instance to test, so those are totally untested. I mostly tested Plex. Also I added the existing weather module as an optional station itself. This is the claude generated description:

## Summary

https://github.com/user-attachments/assets/9b5ebfa5-d025-426a-a0c1-12c65890fcd2

https://github.com/user-attachments/assets/cd2d0cf9-ac1f-4d7e-8a40-6985ccacff5c

Adds a **Channels** module: scheduled linear TV. A channel is built ahead of time into a timeline of programs, commercials, bumps, intros and outros; tuning in resolves `(channel, now)` to `(item, offset)` and hands off to mpv, landing you part way through a program if that's where the schedule is. Nothing runs in the background — a channel costs nothing until someone tunes to it.

- Programs from Local Files, Plex, Jellyfin or Emby, mixed on one channel
- Pick a whole series, a season, or single episodes — a tick means "this airs" at every level, and ticking one carries up to the levels above
- Breaks from folders of your own clips, with per-source idents so a series can be announced in its own voice
- Guide with the clock and a moving preview of what's on (or artwork, or neither)
- Movie slots — a film at 8 PM on chosen days
- Station logo, channel banner, and a card / black / analogue static while a channel loads — all optional
- Balanced shuffle so a channel's series take turns, or straight play order
- Channel up/down on a remote, bound to the standard codes out of the box and rebindable under Settings → Controls

**Ships off.** Nothing appears on the main menu until it's switched on, and nothing else in the app changes behaviour.

Shared code touched, all additive: `AppCore` gains an optional `actionStatus(QString)` signal, wired by introspection like the existing optional signals; `MpvController` gains volume get/set, a channel-banner OSD call, a `playbackStalled` signal and `forceStop()`; `DisplayHandoff` hands the screen back black rather than Qt's stale framebuffer (this fixes a pre-existing flash of the previous screen) and emits `displayReturned`; small read-only additions to the Plex/Jellyfin/Emby/Weather backends.

`release.yml` now installs Qt Multimedia. It's optional and powers one thing — the guide's moving preview — but without it every released build silently falls back to a still card. CMake detects it and prints which way it went.

Plex playback uses Plex's own transcoder when needed, at the quality set in the Plex module, with the session started at the join offset.

Known limitations, stated up front:

- Within a long run of short interstitials a channel can drift up to ~10s behind its own timeline, then re-syncs at the next program. The visible effect is the last break being clipped, which is what a real channel does, so it's left alone.
- Plex occasionally 404s the first segments of a freshly started transcode at a program boundary; mpv skips them and the opening seconds are lost. Seen twice in fifteen boundaries.
- A keyboard's PageUp/PageDown doesn't change channel *during playback* — the Qt window isn't OS-focused while mpv is fullscreen, so those keys reach mpv. Remote and gamepad work, because those are posted to the window.
- ARCHITECTURE.md documents the new `submenu` setting type, but has no section on this module's internals. Happy to write one.

## AI involvement

Written with Claude Code. Essentially all of the code is AI-generated, from my direction on what the feature should do and how it should feel — a long back-and-forth over several days, not one prompt.

Before submitting I ran it on real hardware throughout rather than only at the end, drove the feature decisions and rejected plenty of what came back, and reviewed the result — including an adversarial pass over the whole branch that found and fixed real bugs: a rename flow that had never worked, a decision the playback path computed and then discarded, an item that could replay itself forever, and a stalled stream that wedged a channel indefinitely. I watched multi-hour unattended soaks and chased what they surfaced.

I own this code and can explain any of it.

## Testing

726 automated checks over the schedule format, the tuner decision table, the generator, the media-server client and the path handling. Off by default, no extra dependencies:

```bash
cmake -B build-tests -DBUILD_VCHAN_TESTS=ON . && cmake --build build-tests --target vchan-tests
./build-tests/vchan-tests
```

This repo has no tests today, so there was no existing style to follow — the suite is a ~40-line header of counters linking only Qt6::Core, rather than pulling a framework into a project that has none. Qt6::Test is part of qtbase and already on every machine that builds this, so switching would cost no dependency either. Say the word and I'll convert it, or drop the tests from this PR. There's no PR CI here, so they don't run automatically; happy to wire that up too.

Also tested by hand over several days: every screen reached with remote-only input, and multi-hour unattended soaks — the longest 8½ hours on one channel, with zero off-air moments, zero playback-guard firings, and every program starting within 1–15s of its half-hour mark.

- Platform(s) tested: Raspberry Pi 4 (Raspberry Pi OS), against a real Plex server. Not tested on macOS.
- Display / output tested: CRT over composite (720x480) and an HDMI panel.

## Checklist

shot-3-module-settings
shot-2-channel-settings
shot-1-manage-channels

https://github.com/user-attachments/assets/35109d42-74bb-4186-801a-c7804032e060

https://github.com/user-attachments/assets/92891083-b506-453d-9c4d-abe32304ba55

https://github.com/user-attachments/assets/faebe200-7bd1-4401-89a1-b405790f408c


AchillesPDX:
Oh my good god this looks amazing. Fingers crossed this gets merged in.

anthonycaccese:
Thank you @munderwoods, i'll try to dig into this later in the week (will try for earlier but I am booked on some other tasks currently)

munderwoods:
I found a few issues with the local files channel creation so I fixed those and did a reactor and tried to unify it with the Plex experience more. UX is definitely my week point but hopefully it gives you easy access to get started fast and the ability to mine down into details if you prefer. Local files have a required file structure now for series and movies, similar to Plex etc.

anthonycaccese:
I should get space to dig into this more by the end of this week, ok if I set this to draft state considering the updates you are adding (its more than ok that you are, just thinking to make sure its handled as a draft till then)

munderwoods:
Oh yeah that's fine. I was thinking actually about adding programming slots since there can be issues a bunch of the same series piled up when you're airing episodes in order. Also, I'm going to make a separate PR for flirc compatibility.

munderwoods:
By the way, would you want me to make a pi image builder so you can just hose an image file people can flash to their pi that has everything pre-installed? I already have a system set up for that using packer that I can just port over real quick.

anthonycaccese:
> would you want me to make a pi image builder so you can just hose an image file people can flash to their pi that has everything pre-installed

No thank you, I def appreciate the offer but I'd prefer keeping things app focused in this repo.  It helps keep my scope for helping with things out of the OS side of things.  My feeling is there would be more variables to watch for on OS set ups then I would be interested in directly handling, aiming to keep this more hobby/tinkering focused because of that.

Virtual Channels wise, I started playing with this today on my Mac and i'm having fun but will def need some time to sit with it and wrap my head around everything you've built for it =).   Will look to write up some initial feedback and questions this weekend.

munderwoods:
Great I just pushed up some improvements to make the scheduling better. There are now three scheduling modes. I'll want to add a way to make movie channels and still may add a way to do individual programming blocks.

munderwoods:
Added a movie channel type. I don't think I'll do programming blocks for now mostly because I probably won't use it. I'll just do bug fixes as I find them I think.

munderwoods:
Ok I was wrong. Programming blocks are in. They are a timing option. Lots of ui fixes and improvements in here too that should bring the settings closer in line with the rest of the app.

munderwoods:
I consider this feature complete now. I have been bashing my head against subtitles and audio tracks and finally have it where I want it in plex. This will make the experience much more consistent. You can now select a preferred audio and subtitle language, and you can set subtitles to only display when the only audio track does not match your preferred audio language. Also this fixes an issue where selecting a language for audio or subtitles would sometimes restart the video being played which really takes you out of the channel surfing experience. It's possible this affected the main plex module as well. 

There are lots of other little bug fixes and features in here as well. I'm going to keep using it for the next week or so to look for bugfixes.

indyjamesb:
Hey @munderwoods — thanks for this, I’ve been testing Channels on a Pi 4 against current `main`.

Two small UX issues I hit, both with fixes on a fork test branch if useful:

**1. Weather channel number collision**  
Guide defaults to `0`, Weather to `1`. If you create a user channel while Weather is still off, it also gets `1` (`nextFreeChannelNumber` only reserves Weather’s number when Weather is already enabled). Turning Weather on later leaves two dial entries on `1`, and `move_channel` then refuses to run until they’re apart.

Fix approach: when Weather is turned on (and on startup if it’s already colliding), move the special to the lowest free dial number via the existing `setSpecialNumber` path.

**2. Back during playback leaves Channels**  
From the player (and Weather channel view), Back / stop was calling `exitModule()`, which drops you on the app home instead of the guide. Switched those paths to `goBack()` so Back returns to channel selection.

Fork branch with both fixes (plus #276 merged onto current `main`, including the Plex NFC / `transcodeQuery` conflict resolve):  
https://github.com/indyjamesb/240-MP/pull/1  

Relevant commits:  
- dial collision: `f6a7ba6`  
- Back → guide: `3e3d6dc`  

Happy to open a small PR against your branch, or you can lift those two commits. Testing was AI-assisted (Grok Bot) under my direction; I validated Back-to-guide on hardware. Weather reassignment is in the same branch if you want to try the enable-after-create path.

PR #279, opened 2026-08-31
feat(plex): add browse by folder to the library menu

cookandy:
## Summary

Adds a `FOLDERS` entry to the per-library menu, mirroring browse-by-folder in the Plex web client. It lists the library's on-disk directory tree. Folders drill down, files open the normal detail screen. Useful for content without metadata (such as WOC recordings).

Backed by a new `load_folder` / `folderLoaded` pair on `PlexBackend` hitting `/library/sections/{id}/folder`, following the same shape as the other browse calls. `Items.qml` handles the new `folders` listType.

A couple notes:

- The `FOLDERS` row is unconditional, like `COLLECTIONS` and `PLAYLISTS`
- The A–Z panel assumes an alphabetical listing, so in a folder mixing subfolders with loose files it can jump backwards (Plex returns folders first, then files, each sorted separately)

pi
pi2

## AI involvement

The implementation was AI-generated using Claude Code. I reviewed the diff, built from source on a Pi 3, and and verified the behaviour by hand by browsing folder listings in my own library against a live PMS.

## Testing

Built from source and run on the Pi against a live Plex Media Server. Navigated into a library, opened `FOLDERS`, drilled through directory levels, walked back out with esc, and used the A–Z panel. Played content from top-level and nested directories.
- Platform(s) tested: Raspberry Pi 3B, built with `cmake --build build`
- Display / output tested: HDMI and composite

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Watch on YouTube: https://youtu.be/r-gylGDoELY
 - Autoplay next episode in a season (optional, off by default)
 - Write an NFC card for any movie, episode, season or show from its detail screen to use with the NFC Reader module
 - Hub, Playlist, Collection and Category support
+- Folder browsing — walk a library by its on-disk folder structure
 - Movie editions
 - Select preferred audio/subtitle track before playback and switch tracks during playback
 - Full library browsing by letter
--- a/modules/plex/views/Items.qml
+++ b/modules/plex/views/Items.qml
@@ -18,12 +18,17 @@ FocusScope {
     property string ratingKey: navParams.ratingKey || ""
     property string categoryKey: navParams.categoryKey || ""
     property string libraryName: navParams.libraryName || ""
+    // Folder browsing: the key of the folder being listed (empty at a section's
+    // top level) and its name, which stands in for the library name in the header
+    // so a deep folder still says where it is.
+    property string folderKey: navParams.folderKey || ""
+    property string folderName: navParams.folderName || ""
 
     property var items: []
     property bool isLoading: false
     property string errorMessage: ""
 
-    property bool showLetterNav: listType === "library_all"
+    property bool showLetterNav: listType === "library_all" || listType === "folders"
     property bool letterNavActive: false
     property var letterIndex: []
 
@@ -185,6 +190,20 @@ FocusScope {
             }
         }
 
+        function onFolderLoaded(loadedItems) {
+            if (itemListRoot.listType === "folders") {
+                itemListRoot.isLoading = false
+                itemListRoot.items = loadedItems
+                if (itemListRoot.showLetterNav)
+                    itemListRoot.letterIndex = itemListRoot.buildLetterIndex(loadedItems)
+                if (loadedItems.length > 0) {
+                    var restore = (navListState.currentIndex !== undefined) ? navListState.currentIndex : 0
+                    itemList.currentIndex = Math.min(restore, loadedItems.length - 1)
+                    itemList.positionViewAtIndex(itemList.currentIndex, ListView.Contain)
+                }
+            }
+        }
+
         function onCategoriesLoaded(loadedItems) {
             if (itemListRoot.listType === "categories") {
                 itemListRoot.isLoading = false
@@ -220,6 +239,20 @@ FocusScope {
                 queueItems: queueItems,
                 shuffle: item.__action === "shuffle",
                 title: listTitle,
+                libraryName: libraryName
+            }, { currentIndex: itemList.currentIndex })
+            return
+        }
+
+        // A folder row opens the next level down. The server's own key is passed
+        // back untouched, so the section id only matters at the top level.
+        if (item.type === "folder") {
+            itemListRoot.navigateTo("Items.qml", {
+                listType: "folders",
+                title: item.title,
+                sectionId: sectionId,
+                folderKey: item.folderKey,
+                folderName: item.title,
                 libraryName: libraryName
             }, { currentIndex: itemList.currentIndex })
             return
@@ -326,6 +359,8 @@ FocusScope {
             plexBackend.load_categories(sectionId)
         else if (listType === "category_items")
             plexBackend.load_category_items(sectionId, categoryKey)
+        else if (listType === "folders")
+            plexBackend.load_folder(sectionId, folderKey)
         else if (listType === "continue_watching")
             plexBackend.load_continue_watching()
     }
@@ -346,7 +381,7 @@ FocusScope {
     AppBar {
         iconSource: moduleRoot.moduleIcon
         title: moduleRoot.moduleName
-        subtitle: libraryName
+        subtitle: folderName !== "" ? folderName : libraryName
         anchors.top: parent.top
         anchors.left: parent.left
         anchors.topMargin: root.sh * 0.125 //60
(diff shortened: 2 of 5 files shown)


anthonycaccese:
thank you @cookandy! browse by folder is def a use case I missed, thank you for considering it and adding it. i'll make time to review this when I am back mid week.

anthonycaccese:
Hi @cookandy have been using this today and its great, thank you.

First, I am glad you called out the letter nav because in my library its feels odd when using it with folders =)  

My movie library on disk is structured by letter like this:

```
Movies/
    1/
        10 Cloverfield Lane (2016)/
        12 Monkeys (1995)/
    A/
        Akira (1988)/
        An Inconvenient Truth (2006)/         ← filed under A on disk
        Apollo 13 (1995)/
    B/
        A Beautiful Mind (2001)/              ← filed under B on disk
        Back to the Future Part II (1989)/
```

In the `A/` folder the letter nav displays A and I, because it strips `An` from `An Inconvenient Truth (2006)` and buckets it under I.  

This happens because the current letter nav logic strips leading articles for it to work with library browsing.  But in folder browsing its clear the order should come from the name on disk instead, so the two end up disagreeing.

As a side note: Its not lost on me that my own filing isn't consistent about articles...  `A Beautiful Mind` is under B but `An Inconvenient Truth` is under A.  Something I should likely fix now that I can't unsee it =)

Separately (and I haven't been able to reproduce this myself since my tree is folders all the way down): if a folder contains both subfolders and loose files, do the folders always sort ahead of the files? If so, a letter that appears in both would only ever jump to the folder.

So I wanted to confirm, is one of your goals for this to...
"respect and display (without any additional processing) where I have things stored and how they are named on disk"

If yes, what do you think about dropping the letter nav for folder browsing?

The downstream impact would be cases where users have a more flat hierarchy higher up in the root tree (like TV/[List of shows]). I know a browse by letter could be helpful for that.  

Ultimately I think some of the quirks with letter browsing for folders are likely ok to accept so I would also be ok with keeping it. 

But as a third option for folder browsing only... we could look at bucketing folder rows by their literal first character rather than the article-stripped one and then only show the letter nav when the result makes sense to display it.  Specifically: more than one bucket, a single alphabetical run, and not a folders-plus-files mix. That could be easy to add and it'd work for a flat `TV/[shows]` root while staying off when browsing an `A/` folder.

For your use case of "content without metadata (such as WOC recordings)" how does browsing via the "TV Shows (WOC) > Library" work for you in practice for that kind of setup?

Thank you for your feedback, it will help me think this through a bit further.

---

Second, on the app bar subtitle...

At the top FOLDERS level it shows the library name, and from the first child down it swaps to the folder name. 

I saw the comment that you added the folder name so "a deep folder still says where it is" which makes sense. 

But i'm not sure it fully gets there since only the leaf name is shown. Browsing to Movies/A/Akira (1988)/ gives AKIRA (1988) with nothing saying which library I'm in, and Season 1 folders from 2 different shows would give identical headers of just "Season 1"

What do you think of something like `MOVIES / AKIRA (1988)` (i.e. keeping the library and appending the current folder)?

cookandy:
Thanks for your feedback!

IMO folder browsing should respect and display where things are stored and how they're named on disk. I don't think it should be the way you browse a well-tagged movie library (LIBRARY and CATEGORIES already do that better anyway). In my mind, folder view is for content without metadata where the folder structure is the organization.

So I'd probably prefer the third option. Bucketing on the first character seems correct, because article-stripping is metadata logic leaking into a view that's supposed to be about filenames. It also fixes your case, inside `Movies/A/` every entry starts with `A`, so there's a single bucket and the panel hides itself; at `Movies/` root you get `1`, `A`, `B`, etc.

I'd personally like to keep the letter nav rather than drop it since I have folders with a lot of items, and paging through them with up/down is rough.

On when to show it, I think your three checks collapse into one with the third option: show the panel only if the list is already in alphabetical order and has more than one starting letter. That handles the mixed case for free. A folder with subfolders A, B, C and files A, B, C will be A, B, C, A, B, C, so the alphabet restarts, the list isn't in order, and the panel switches off. And a flat `TV/[shows]` root still gets it. Today a letter appearing in both only ever jumps to the folder. This rule should make the panel hide itself in those cases.

On the app bar, agreed, leaf-only isn't enough, and `MOVIES / AKIRA (1988)` is better. One small issue is with TV shows. Once you're in a show, `Season 1` folders will look identical, since only the library and the leaf would be shown. If we want those distinguished, we'd need the parent in the breadcrumb too, which should be possible.

On your last question, LIBRARY doesn't really work for me with the WOC recordings. My top-level folders are networks, and inside each the files are normally named `YYYY-MM-DD - Network - Show.mp4`. The library view sorts alphabetically, which for these filenames means chronologically, so the networks all interleave (`1999-01-01 - ABC - Saturday cartoons.mp4` is right next to `1999-01-01 - Nickelodeon - Snick.mp4`), etc. It's harder to browse for a specific show from a specific network. 

Saying that, I guess the letter navs won't help a ton with my filenames, but I think it'd still be useful to keep them for folder browsing (or for folks who might organize their files differently). 

Curious to hear your thoughts!

anthonycaccese:
> IMO folder browsing should respect and display where things are stored and how they're named on disk. ... In my mind, folder view is for content without metadata where the folder structure is the organization.

Dig it, I agree

> So I'd probably prefer the third option. Bucketing on the first character seems correct, because article-stripping is metadata logic leaking into a view that's supposed to be about filenames

Awesome, i'm with you on keeping it and trying out option 3

> On when to show it, I think your three checks collapse into one with the third option: show the panel only if the list is already in alphabetical order and has more than one starting letter. That handles the mixed case for free

That sounds great, is it something you'd be interested in adding? or I am glad to work on it as part of this PR as well just let me know what you prefer.

> On the app bar, agreed, leaf-only isn't enough, and MOVIES / AKIRA (1988) is better. One small issue is with TV shows. Once you're in a show, Season 1 folders will look identical, since only the library and the leaf would be shown. If we want those distinguished, we'd need the parent in the breadcrumb too, which should be possible.

Yeah good catch, I would be down for full breadcrumb too.  The appBar is built to elide right so it will naturally cut off longer subtitles with an ellipse but hey we're talking about displaying this on a CRT so I think its ok and expected =)

> LIBRARY doesn't really work for me with the WOC recordings. My top-level folders are networks, and inside each the files are normally named YYYY-MM-DD - Network - Show.mp4

Thank you for writing this out, its a cool use case I hadn't run into and def helped me understand better the benefit of folders for that kind fo set up.

cookandy:
> That sounds great, is it something you'd be interested in adding?

No problem! I'll try to spend some time on it this week, and will tag you when it's ready for review. Cheers

cookandy:
@anthonycaccese  I pushed both changes if you want to test. The folder letter nav now buckets on the literal first character and only appears when the list is in order with more than one letter, and the app bar shows the full trail.

One thought for later: long trails elide on the right, so the deepest folder is the part that gets cut. It might be worth scrolling it like the list rows, or eliding the middle instead. Wdyt?

anthonycaccese:
Awesome! Your changes look great @cookandy and I think this is in a great state to merge.

> One thought for later: long trails elide on the right, so the deepest folder is the part that gets cut. It might be worth scrolling it like the list rows, or eliding the middle instead. Wdyt?

Yeah good call and I agree it would make sense as a follow up later.  I am thinking eliding the middle could be interesting to try (thinking that seeing the beginning and end of the string may provide the needed summary to help anchor) so would try that first but if it feels odd in practice scrolling it would work too.

cookandy:
Excellent, thanks for the review! Feel free to merge when you're ready. :)

Appreciate your feedback and quick turnaround!

anthonycaccese:
thank you right back, stoked for you finding this and having the interest to contribute.  great contribution.

PR #284, opened 2026-09-03
Support IR receivers that present as a single keyboard (Flirc)

munderwoods:
This allows Flirc to work on 240mp, hopefully alongside the recommended type of controller. I have not tested it, as I do not have one, so if someone else could I would appreciate it. With this I was able to get 240mp working with my universal remote (a sofabaton x2) so it's seamless with the rest of my entertainment system. The rest of this PR was written by Claude, and reviewed by me. 

## Summary

An IR receiver that presents as one USB keyboard — a Flirc, and most receivers
like it — sends `KEY_SELECT` for OK and `KEY_CHANNELUP` / `KEY_CHANNELDOWN` for
channel up and down. No keyboard layout has a keysym for those, so EGLFS
delivers a `QKeyEvent` whose `key()` is `0` and the buttons do nothing.

They could not be bound out of it either: the Controls screen captures whatever
it is handed, `0` included, and `0` is what the app stores to mean "not bound" —
so the row went back to reading "default" and the press looked unrecognised. The
recovery path needed the one thing that was broken.

`InputManager` already reads these codes off the device, but only opened a device
whose *name* contains "Consumer Control", which is how a multi-interface remote
exposes them. A single-interface receiver puts them on the keyboard, where that
test never looks. It now falls back to the first device that *advertises* the
codes, whatever it is called.

That device is also the plain keyboard, so only the codes Qt cannot deliver are
taken from it — anything Qt does deliver would arrive twice and fire its action
twice. Gamepads are excluded; they belong to SDL. A dedicated Consumer Control
interface still wins the scan and is still read in full, so existing remotes are
unaffected.

Also: `KEY_SELECT` gets `Action::Select` by default, and a capture the app cannot
store is now refused on screen instead of silently saved as `0`.

Channel up/down are read and become bindable here. They act once the actions they
map to exist, which is a separate change.

## AI involvement

Written with Claude Code (Opus). The diagnosis came from captured `evdev` events
off the receiver, which is what identified `KEY_SELECT` rather than `KEY_ENTER`
and the `key() == 0` path; the implementation is AI-authored. I reviewed the
change and tested it on hardware with the actual remote before submitting —
see Testing below for what was exercised directly and what was simulated.

## Testing

- Platform(s) tested: Raspberry Pi 4 Model B, Raspberry Pi OS Trixie (Debian 13),
  Qt 6.8.2, EGLFS. Not tested on macOS — the changed path is `Q_OS_LINUX` only.

With a Flirc and its remote:
- Select, Channel Up and Channel Down now work; before the change all three did nothing.
- Arrows and other ordinary keys still fire once per press (the double-handling risk).

Original multi-interface remote path, simulated with a `uinput` device named
"… Consumer Control" alongside the Flirc:
- The dedicated interface is still chosen even though the Flirc is scanned first
  and matches the new fallback, and the original log line is what appears.
- Removing that device mid-run tears down cleanly (existing disconnect path).

Not exercised on device: the new "cannot be remapped" message in the Controls
screen — the guard is straightforward but I have not seen it on the TV.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ Thumbs.db
 # 240-MP #
 ######################
 /build
+/build-*
 /.cache
 # AppImage build outputs (scripts/build-appimage.sh, scripts/build-mpv.sh)
 /AppDir
--- a/src/input/InputManager.cpp
+++ b/src/input/InputManager.cpp
@@ -381,6 +381,12 @@ void InputManager::loadKeyRemap() {
         { "select", Action::Select },
         { "back",   Action::Back },
     };
+#ifdef Q_OS_LINUX
+    // A remote's OK button sends KEY_SELECT, not KEY_ENTER. Seeded before the
+    // stored bindings so a viewer's own choice still wins.
+    m_keyRemap[kEvdevKeyBase + KEY_SELECT] = Action::Select;
+#endif
+
     for (const auto &entry : kRemapActions) {
         bool ok = false;
         const int qtKey = m_appCore->get_setting(QString(), QStringLiteral("remote_keymap.") + entry.name).toInt(&ok);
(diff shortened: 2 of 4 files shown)


anthonycaccese:
@munderwoods, I am not in a place where I can look at the changes for this yet (should be back home tomorrow) but does this PR essentially add the ability to be able to use a FLiRc USB recevier without first mapping it using their software on a PC?   

Asking because I have a Flirc usb receiver and it worked well with 240-MP but I had done the mapping on my mac first using their software to standard keyboard controls

munderwoods:
Hmm, is it a Flirc 1 or 2? Mine is the two, and perhaps that is the difference. The select and channel buttons definitely did not work out of the box.

anthonycaccese:
I have this model: https://flirc.tv/products/flirc-usb-receiver?variant=43513067569384

I used their mapping software on my mac to map my IR remote keys to keyboard keys for up/down/left/right and select back then plugged in the recevied to pi and it worked directly without any additional mapping

munderwoods:
Hmm, maybe I will revert mine to stock and try again then. I'll let you know!

Edit: To be clear, you just remap the ir signals it's listening for, correct? To my knowledge, the signals coming out of the flirc are always the same unless you get pretty advanced.

munderwoods:
> I have this model: https://flirc.tv/products/flirc-usb-receiver?variant=43513067569384
> 
> I used their mapping software on my mac to map my IR remote keys to keyboard keys for up/down/left/right and select back then plugged in the recevied to pi and it worked directly without any additional mapping

Oh, do you know which config you used in the Flirc GUI?

anthonycaccese:
yup most def, i just used their desktop app directly 

thought it might help to see a video of how its set up on my end:

https://github.com/user-attachments/assets/6e7f0f4b-c5b5-4785-9f28-5adf8024dfc5

> Oh, do you know which config you used in the Flirc GUI?

nothing that specific for me, I essentially just opened their gui, plugged in the receiver and then used their in-app buttons to map each button on my ir remote

anthonycaccese:
hi just wanted to check in to see what you found after testing and my video above.  I am away for a bit longer but plan to be back by end of the week.

PR #292, opened 2026-09-10
feat(dlna): add DLNA/UPnP media server module

ScarletPachydermDev:
## Summary

Adds a DLNA/UPnP module: discovers MediaServers on the local network over SSDP,
browses their ContentDirectory, and hands off to mpv for playback via the existing
`MpvController`.

Implemented with Qt's own networking and XML — no new libraries. The module is
self-contained under `modules/dlna/` and `src/modules/dlna/`; the only shared-code
change is one line in `CMakeLists.txt` and three in `src/main.cpp` (include,
construction, `registerModule`). 11 files, +2258 / -0.

A server's top level is usually Browse Folders / Music / Pictures / Video. Since the
folder tree already contains all of it, the module opens the folder view directly
rather than making the user pick a category first; servers exposing no such container
fall back to showing whatever they do have.

Features: folder browsing with up-navigation, resume positions persisted to
`dlna_history.json`, pinned main-menu shortcuts via `get_menu_entries()`, sidecar
subtitle loading, and optional `ffprobe`-based track-language listing. Ships with
`enabled` defaulting to `OFF`, so existing installs are unchanged.

## AI involvement

**The code and this description were written by Claude** (Claude Opus 5, via Claude
Code). Not AI-assisted — AI-authored, end to end. I directed it, made the product
decisions, reviewed the source, and tested it, but I did not hand-write the
implementation.

What I did:

- Set the requirements and made the design calls (what the module should do, key
  bindings, which settings to expose, what to cut).
- Tested every iteration on real hardware against a live DLNA server, and reported
  defects back. Most of the significant bugs were found this way rather than by
  reading the diff.
- Read through the source line by line, in addition to testing behaviour.

Worth knowing for your review: the protocol layer was wrong in several ways on the
first pass and only became correct through testing against a real server. Claude had
invented API signatures rather than reading them — the SOAP action was being POSTed
to `SCPDURL` instead of `controlURL`, the DIDL-Lite payload inside `<Result>` was
never unescaped, element matching used `dc:title` where `QXmlStreamReader` reports
local names, and `<res>` attributes were read after `readElementText()` had already
advanced past them. All are fixed and verified working, but **the SOAP/DIDL handling
is the part I'd most want a second pair of eyes on**, since it was arrived at
empirically against one server.

Also flagging: the repo's `CLAUDE.md`, `CONTRIBUTING.md` and this PR template were
only read *after* the PR was first opened. Two violations turned up on that pass and
are fixed in the branch — hardcoded `radius: 4` pixel values, and `navListState`
position restoration being declared but never honoured.

## Testing

Manual, against **MiniDLNA 1.3.x**: discovery, nested folder browsing, up-navigation,
playback, resume across restarts, pinned shortcuts, sidecar subtitles, and track
probing. Verified with AAC, TrueHD 5.1 and E-AC-3 5.1 audio. Confirmed that
renderers advertising no ContentDirectory (an NVIDIA Shield) are correctly skipped
rather than listed as broken servers.

- **Platform(s) tested:** Steam Deck (SteamOS), building inside a Fedora 44
  distrobox. **Not tested on macOS or Raspberry Pi OS** — I don't have either set up,
  so the two documented build targets are unverified.
- **Display / output tested:** 4K HDMI (3840×2160). **Not tested on a CRT at
  240p/480i.** Layout uses `root.sh`/`root.sw` throughout and reuses the same margins
  and offsets as the Local Files module, so it should inherit that module's safe
  area — but I can't claim it's been seen on the project's primary target.

Not tested against any other server (Plex DLNA, Windows Media Player, Serviio,
Jellyfin). The SOAP and DIDL handling follows the spec rather than MiniDLNA quirks,
but given how it was arrived at, a second server would be worth checking before merge.

## Checklist

Notes on the checklist:

- Navigation was exercised with a keyboard and the Deck's controller, not a physical
  USB remote. Every action maps to the allowed key set — Right pins a server/folder
  and plays-from-start on the detail view, matching the scripts and YouTube modules'
  d-pad rationale.
- CRT overscan is inherited from Local Files' margins rather than independently
  verified, per the display note above.
- The only outside network traffic is to the user's own LAN DLNA server. State is
  two JSON files in the data directory (`dlna_pins.json`, `dlna_history.json`),
  following `local_files_history.json`'s read-modify-write shape.

## Known limitations

- **No search**, even where a server advertises the capability.
- **Sidecar subtitles depend on the server advertising them.** A `.srt` next to the
  video that the server doesn't expose in its DIDL response can't be discovered over
  DLNA at all. Muxed tracks are likewise invisible to DLNA and surface only through
  the `ffprobe` path.
- **`ffprobe` is an optional dependency**, used only to list track languages. Invoked
  via `QProcess` with a 10 s guard; a failure to launch is caught and the detail view
  falls back to the channel count DLNA reports. Flagging it since Pi and AppImage
  builds may not ship it.
- **`--alang` is passed through `extraArgs`**, since `loadAndPlay()` has no audio
  language parameter. Reaching `extraArgs` means spelling out all 16 preceding
  positional arguments, which is fragile if that signature ever changes. A
  first-class parameter alongside `subLangs` would be cleaner, but that's a
  shared-code change I didn't want to make unasked.

## Notes for review

- Module state (folder trail, current container) lives on `moduleRoot`, not in
  `BrowseView`, because opening the detail view unloads that view and it's rebuilt
  from its original `navParams`.
- Views follow the house conventions: `FocusScope` + `AppBar`, list anchored to
  `parent.top` at a fixed offset, white icon fill, `root.hints.*` footer labels
  (never `inputManager.hints.*`), and `navigateTo`/`goBack` signals rather than
  calling router functions directly.

---


anthonycaccese:
@ScarletPachydermDev, I appreciate the PR and i'll review it soon.  Ahead of that is it safe to say the PR message above was fully generated? (I am making that assumption based on the `generated with Claude Code` note at the bottom)  If yes, please can you confirm and write out your review steps for it and add in AI Usage details?  Thank you.

ScarletPachydermDev:
Sorry yes — confirming plainly: the PR description was fully generated, and so was the
module itself. Claude Opus 5 via Claude Code wrote all of it end to end. I directed
it and tested it, but I didn't hand-write the implementation.

Apologies for the missing disclosure — I opened the PR with a custom body and
bypassed your template, so the AI involvement section never got filled in. That's on
me. I've rewritten the description to follow `.github/pull_request_template.md`
properly, with a full AI involvement section, the testing details, and the checklist.

My review steps, honestly stated:

1. **Behaviour testing on hardware, every iteration.** Steam Deck against a live
   MiniDLNA server — discovery, nested browsing, up-navigation, playback, resume
   across restarts, pinned shortcuts, sidecar subtitles, track probing. Verified
   across AAC, TrueHD 5.1 and E-AC-3 5.1.
2. **Reporting defects back and re-testing until fixed.** Most of the real bugs were
   caught this way, not by reading the diff — playback silently failing, back
   navigation skipping a level, text overflowing the screen.
3. **Product decisions were mine** — key bindings, which settings to expose, what to
   cut, the "no dedicated key, use the OSC" outcome for audio/subtitle switching.
4. **Read the source line by line**, not just the behaviour.

The limits that remain aren't about review effort — it's that the protocol layer has
only ever been exercised against one server, on one platform.

Two things I'd point you at specifically:

- **The SOAP/DIDL layer is where I'd focus.** It was wrong in four separate ways on
  the first pass and only became correct through testing against a real server —
  wrong endpoint (`SCPDURL` instead of `controlURL`), the `<Result>` payload never
  unescaped, `dc:title` matched where `QXmlStreamReader` gives local names, and
  `<res>` attributes read after the reader had advanced. All fixed and working, but
  it was arrived at empirically against exactly one server.
- **It's never run on your actual targets.** Not macOS, not Raspberry Pi OS, and not
  on a CRT at 240p/480i — only a Deck on 4K HDMI. Layout uses `root.sh`/`root.sw`
  throughout and reuses Local Files' margins, so it should inherit that safe area,
  but I can't claim it's been seen on a CRT.

I also read `CLAUDE.md` and `CONTRIBUTING.md` only after opening the PR, which turned
up two violations of your own rules — hardcoded `radius: 4` pixels, and
`navListState` position restoration declared but never honoured. Both are fixed
locally; I'll push them once I can re-test on hardware rather than sending you
untested changes.

Happy to cut scope if that helps review — the pinned-shortcuts and ffprobe
track-language pieces are separable from the core browse-and-play if you'd prefer a
smaller first PR.

